PR #5, opened 2018-03-26
create empty structure

a-maur:


apuignav:
Argh! You're creating a huge diff by renaming the file and then adding it.

Can you do `git mv` for the files in `core`?

PR #6, opened 2018-03-27
Try again with empty structure

a-maur:
I'm sorry for messing up names, I don't know how to fix if.
I promise that my current fork is identical to the master (for files in core/)


apuignav:
OK, no worries, the problem is that the commit that created the problem is quite far away, so it'll be hard to fix. Let's merge and that's it!

PR #7, opened 2018-03-27
Merge WIP into zfit-branch

a-maur:
files contains all the implementation of the physics, need to be structured... :)


apuignav:
@a-maur, @rsilvaco I was not in this discussion: what are the plans for the structure of physics? Right now, with this implementation they are not importable and if we want to make the physics "discoverable" we need a bit more structure. Can you summarize how you want to organize things?

mayou36:
@apuignav so we just discussed a possible structure. We have some ideas and do a structural implementation tomorrow, it's easier than to describe it here.

apuignav:
OK! But this PR is already in the master (I don't know what happened). Can you check?

mayou36:
Hm, I don't think it is already in the master of zfit but from andreas master. Right?

apuignav:
You're right, my bad. I must have misread something.

PR #16, opened 2018-04-09
Require >= instead of == for requirements

mayou36:
Makes dependency handling easier overall (no conflicting deps). And it's usually not required to have a ==.


apuignav:
Good idea!

PR #24, opened 2018-09-18
CI setup

mayou36:


apuignav:
Question: do we want to support all these `3.x` python versions?

mayou36:
Well, I'd aim for 3.5, 3.6 and 3.7, 3.4 is though still quite around. For the development, let's keep as many as possible to see if anything breaks (and if so, then decide to may drop any), and before publishing the first "stable release", we gonna decide what to fully support and what not for how long (still dreaming of dropping 2.7 :D). Probably, 3.4 will still be wanted. Agree?

(But we may can drop some integration on smaller branches to save resources)

apuignav:
OK, fair enough, let's see if we need anything that is not compatible with the older versions...

mayou36:
Changes (FYI @apuignav):
- removed 3.5 from CI (if 3.4 and 3.6 -> 3.5 will works as well...)
- added auto-deployment to pip
- added coverage report

Merging (non-critical)

PR #33, opened 2018-10-24
base pdf implementation

mayou36:
This is a more or less finished implementation of the base pdf class, which allows to inherit from and create your own pdfs.

In the "examples" folder, a Jupyter Notebook (simple_dist1) is given demonstrating the basic capabilities. Check it out if you like @apuignav @rsilvaco and let me know what you think.

On the internal structure for the user:
easier usecase:
 - overwrite _publicfunc(), don't care about the yield (and possibly get rid of normalization issues with the `@no_norm_range` decorator)

advanced usecase:
 - overwrite `_hook_publicfunc()` and have full control over the input and output. Just call inside the `super()._hook_publicfunc()`

Everything else of the structure is to make it easier inside the pdf

What's missing:

- [x] cleanup yield handling (currently WIP)
- [x] unittests
	- [x] range
	- [x] analytic integral selection
	- [x] basePDF: ?
- [x] docs everywhere
	- [x] range module
	- [x] range class
	- [x] basepdf module
	- [x] basepdf class
	- [x] integrate module
	- [x] integrate class
	- [x] pdfs: basic
	- [x] pdfs: functor
	- [x] pdfs: dist_tfp
	- [x] pdfs: special
- [x] example on advanced (with own pdf) usage -> later
- [x] cleanup analytic integral implementation
- [x] implement auto multiple limits handling

Probably:
- [x] (improved sampler using inverse analytic integral (icdf))

Later on, more PDFs can be implemented (gaussian, CB etc.)


zfit/core/minimizer.py
```diff
@@ -4,3 +4,266 @@
...
+        """
+        values = [p.read_value() for p in parameters]
+        # TODO: implement if initial val not given
+        return values
+
+    @tolerance.setter
+    def tolerance(self, tolerance):
+        self._tolerance = tolerance
+
+    def minimize(self, loss, var_list=None):
```
apuignav: Quick question: minimiser doesn't evaluate errors?

mayou36: Not yet, but will. I mostly focused on the basepdf here, the minimizer is a fast-setup thing to have something work together.

As you asked here anyway: A better name for `loss`? `cost`?

mayou36:
@apuignav except of additional unittests and docs everywhere, the branch (most important the basepdf.py) is ready. You wanna have a look?

zfit/core/limits.py
```diff
@@ -0,0 +1,476 @@
...
+    ANY_LOWER = object()  # TODO: need different upper, lower?
+    ANY_UPPER = object()
+    __HASH_DELIMINATOR = object()
+
+    def __init__(self, *, limits=None, lower=None, upper=None, dims=None,
+                 convert_none=False):  # TODO: change lower, upper?
+        """Range holds limits and specifies dimension.
+
+        Args:
+
```
apuignav: Missing Args

zfit/core/limits.py
```diff
@@ -0,0 +1,476 @@
...
+            convert_none (): If True, None will be converted to "any" (either any lower or any upper)
+        Returns:
+            zfit.core.limits.Range:
+        """
+        # TODO: make use of Nones?
+        return Range(lower=lower, upper=upper, dims=dims, convert_none=convert_none)
+
+    @classmethod
+    def from_limits(cls, limits, dims, *, convert_none=False):
+        """            limits (tuple): A 1 dimensional tuple is interpreted as a list of 1 dimensional limits
```
apuignav: Something went wrong with the doc here?

zfit/core/limits.py
```diff
@@ -0,0 +1,476 @@
+from __future__ import print_function, division, absolute_import
```
apuignav: This is one of the modules where a general module-level doc, explaining a bit the Range object, would be very nice (although not necessary at this stage).

apuignav:
Yeah! I'll have a look at the PR in detail probably on Wednesday. I've started, but I won't have so much time until then.

zfit/core/limits.py
mayou36: done

zfit/core/limits.py
mayou36: Done

zfit/core/basepdf.py
```diff
@@ -0,0 +1,1097 @@
...
+
+Before we create an instance, we need to create the variables to initialize it
+>>> mean = zfit.FitParameter("mean1", 2., 0.1, 4.2)  # signature as in RooFit: *name, initial, lower, upper*
+>>> stddev = zfit.FitParameter("stddev1", 5., 0.3, 10.)
+Let's create an instance and some example data
+>>> gauss = MyGauss(mean=mean, stddev=stddev)
+>>> example_data = np.random.random(10)
+Now we can get the probability
+>>> probs = gauss.prob(x=example_data, norm_range=(-30., 30))  # `norm_range` specifies over which range to normalize
+"""
```
apuignav: Add blank line between docs and first import?

zfit/core/basepdf.py
```diff
@@ -0,0 +1,1097 @@
...
+
+A simple example:
+>>> class MyGauss(BasePDF):
+>>>     def __init__(self, mean, stddev, name="MyGauss"):
+>>>         super().__init__(mean=mean, stddev=stddev, name=name)
+>>>
+>>>     def _unnormalized_prob(self, x):
+>>>         return tf.exp((x - mean) ** 2 / (2 * stddev**2))
+
+Notice that we only specify the *function* and no normalization (we could but would not change anything).
```
apuignav: I am not sure this sentence is clear. It sort of implies that it is not possible to specify integration?

zfit/core/basepdf.py
mayou36: Yep, done

zfit/core/limits.py
mayou36: Done (will be fully done when the docs automation is set up)

zfit/core/basepdf.py
mayou36: I agree. I extended it and also changed the formulation

mayou36:
Well, there you go. Finished so far. If you have any comments left, let me know, otherwise, approve the merge.

I'm gonna do it locally in order to cleanly resolve the merge conflicts.

apuignav:
OK! So I'll finish my review tonight, and I'll merge.

Albert

mayou36:
Okay, that's good!

On the merging: if you just say that it's ok, I can do the merging, as there are merge conflicts that have to be resolved and are most easily done local (at least for me). But feel free to merge yourself!

zfit/core/basepdf.py
```diff
@@ -0,0 +1,1097 @@
...
+                                     "".format(name=caller_name))
+                else:
+                    norm_range = False
+            else:
+                norm_range = self.norm_range
+        return convert_to_range(limits=norm_range, dims=dims)
+
+    @property
+    def dims(self) -> typing.Tuple[int]:
+        """Return the dimensions (a tuple of integers) of the function.
```
apuignav: This one is a bit unclear... What is it its aim?

zfit/core/basepdf.py
```diff
@@ -0,0 +1,1097 @@
...
+    @property
+    def is_extended(self) -> bool:
+        """Flag to tell whether the pdf is extended or not.
+
+        Returns:
+            bool:
+        """
+        return self._yield is not None
+
+    @staticmethod
```
apuignav: Do we want this to be static and so public? I would think this is only used internally, but maybe I'm missing a use case...

apuignav:
So, I've had a look. I put a couple of comments, but they are super minor. My only worry is that the classes exposes tons of "public" methods, and it may be confusing, so we just need to document things very well and make sure that only a handful are needed in "normal" life.

I will approve this, so you can get down to merging, but the next step will be to work on the docs before adding more features? @rsilvaco Maybe we need to add convolution first?

zfit/core/basepdf.py
mayou36: Basically just to return the axis. So given some n-dim limits, return (0, 1, 2,...) or whatever they have been specified as. Would you prefer 'axis'?

zfit/core/basepdf.py
mayou36: No, you're right, good catch! Is not needed anymore, removed

zfit/core/basepdf.py
apuignav: So, this just returns the tuple of axes? Then let's rename this to `axis`, although if we can remove it let's remove it.

mayou36:
Perfect! About the public methods, in normal production you really only need `prob`, `set_yield`, `integrate` and `sample` (and probably `temp_norm_range`).

apuignav:
> Perfect! About the public methods, in normal production you really only need `prob`, `set_yield`, `integrate` and `sample` (and probably `temp_norm_range`).

Yeah, that was my concern, so much "bagagge". However, here we'll have to write good docs and that's it. I would say I'm happy.

mayou36:
Agree!

And I would finish some features for sure, let's say implement some pdfs (like convolution), finish the Minimizer API and some implementations and get something small with the data working. But of course, docs can also be done underway

zfit/core/basepdf.py
mayou36: Yeah true, in the current status it is not needed. The idea was that you have a pdf with the axis (0, 3), one with (1, 2), but it's really not needed. Removed it.

PR #38, opened 2018-11-06
create Base minimizer and API

mayou36:
WIP!

- MinuitMinimizer works already for minimization
- MinimizerInterface is implemented (`abc.abstractmethod` is mandatory to override)
- BaseMinimizer takes care of the public functions and allows for simple minimizer subclassing
- TF optimizer getting wrapped instead of subclassed

@apuignav @rsilvaco is this what you would have expected?

### DIFF
--- a/zfit/core/minimizer.py
+++ b/zfit/core/minimizer.py
@@ -77,9 +77,12 @@ class MinimizerInterface(object):
 
 
 class BaseMinimizer(MinimizerInterface):
+    _DEFAULT_name = "BaseMinimizer"
 
-    def __init__(self, loss, parameters=None, tolerance=1e-8, name="BaseMinimizer", *args, **kwargs):
+    def __init__(self, loss, parameters=None, tolerance=1e-8, name=None, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        if name is None:
+            name = self._DEFAULT_name
         self._current_error_method = self._error_methods.get('default')
         self._current_error_options = {}
         self._minimizer_state = MinimizerState()
--- a/zfit/core/minimizer.py
+++ b/zfit/core/minimizer.py
@@ -84,8 +84,7 @@ def _raises_error_method(*_, **__):
     raise NotImplementedError("No error method specified or implemented as default")
 
 
-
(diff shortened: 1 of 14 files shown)


zfit/core/minimizer.py
```diff
@@ -5,66 +5,214 @@
...
+
+    @property
+    def edm(self):
+        return self.get_state(copy=False).edm
+
+    @property
+    def fmin(self):
+        return self.get_state(copy=False).fmin
+
+    def hesse(self):
```
apuignav: I guess we can't just remove it? It would break with `abc`?

mayou36: "yes", as long as it is not overriden by the subclass. But why not provide a default implementation?
It's on the TODO ;)

apuignav: It's a philosophy thing: what does a "default" implementation mean? Each minimizer is going to be different, isn't it?

apuignav: Mmm, to be honest, I'm not sure. Maybe let's leave it like it is for now, then we can revisit if the need appears.

mayou36: > like now
-> with a default implementation?

apuignav: Yeah, or if nothing sensible can be made, raise a `NotImplementedError`.

apuignav:
I am not sure I understand these step-wise minimisers: can you clarify?

mayou36:
Sure: it is what is default in TensorFlow optimizers. `step` does one step of the minimization. And e.g. the TF.Optimizers do this.

It's just another way of defining a minimizer. Advantage: you don't have to care about the termination (if you implement just `step`)

mayou36:
Yes, that's the idea. And if you provide _minimize, that's even better

apuignav:
Cool, I'm happy with this, except all the "WIP" part 😉. Do you want me to approve and you do the merge, or you want to clean that up first?

mayou36:
First clean up. Then I'll remove the WIP and ping you again ;)

mayou36:
Looks good from my side, merge?

PR #40, opened 2018-11-07
License

apuignav:
Change license to BSD-3.


mayou36:
I think you want into develop, right? So we can then merge from develop into master (yeah, that's a pity with the default branches in github... but once the releases start, it will make sense)

Speaking of which: I change, while in pre-release development mode, the default branch to develop. So people visiting zfit see the current develop status

apuignav:
Ah yeah, I fucked up. And yes, change the default to develop please.

mayou36:
Done. And merged!

PR #57, opened 2018-11-11
Revert "Revert "WIP: Add gradient to basepdf""

mayou36:
While funny, this seems to be [the way to go](https://stackoverflow.com/questions/27852143/how-to-pr-and-merge-again-after-reverting-pr-using-github-revert-button), as a PR of the original branch does not work (says the commits have already been merged). Learnign stuff every day.

One open question though:
what does gradient mean? so gradient(x, [p1, p2]) is [(x[0]/dp1, x[0]/dp2), (x[1]/dp1, x[1]/dp2), ...]

But honestly, at this points I don't mind, whatever the others like. In TF we don't use it at all. But the function proves: it works :)


mayou36:
But actually, from my side it's ok. Shouldn't be anything more than a little TF demonstration :D

apuignav:
I'm not sure, we should ask what they want 😆. I think typically these gradients are done in 1-D, but not completely sure.

BTW, I've updated https://github.com/zfit/zfit-tutorials/blob/master/API.ipynb to reflect this MR, check that it is OK before merging this one.

mayou36:
Yeah, but whatever it is, it can be provided  ;) That's the least of our problems.
I've made a very few changes to the notebook, but looks really good!

Merging!

PR #62, opened 2018-11-14
Add basic logging

apuignav:
Logger configuration is added.

Pending question: should we add a `to_file` option that enables streaming to file, together with `to_screen`? This would only work for first instantiation, so it could cause confusion when used by a user. A solution to this would be to have a separate method: `log_to_file(logger_name, file_name, log_level)` which would inspect the handlers and add whatever was necessary, and something like `log_to_screen(logger, log_level)` that could be used to hide the printing to screen.

What do you think? @mayou36

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ numpy
 iminuit
 typing
 pep487
+colorlog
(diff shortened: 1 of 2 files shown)


apuignav:
Ready on my side.

PR #81, opened 2019-01-29
Refactor shape of x

mayou36:
Before, we had a shape of (n_obs, n_samples)
now we switch to (n_samples, n_obs)

this means that value[5] return the 6th "event" and is more consistent with numpy behavior

Furthermore, the return value of `pdf` (and func currently) has shape (n_events) (and not (n_events, 1) as before. Rational: pdf return is always 1d -> expect it to be 1d)


apuignav:
Yessss

PR #85, opened 2019-02-01
Add accessor to converged.

apuignav:
This seemed to be missing


mayou36:
Yes you're right of course!
merge

PR #86, opened 2019-02-01
primitive Autodocs

mayou36:
Add some autodocs. We can surely change them, but to have a basis. Deployment to github pages.

@apuignav what do you think? with ./make_docs.sh we can generate the docs. And deploy them automatically with travis to github pages (for the moment being?)

I am absolutely free for any change of theme (which?) or a website that includes this docs somehow. But it could help us also during development to test the docs, graphical display etc.

Or what would you do about it?


mayou36:
Alright, it's something and we sure can edit it all the time to see what we like

apuignav:
Yeah, let's merge and push so we can see them.

Issue #83, opened 2019-01-30
sample should return something fixed


mayou36:
Just notes, will be complited later.
If we return a tensor from sampling, this can lead to sampling again every time the graph is run. Usually, we don't want that. So the sample should be stored in a Variable or similar.

Best would be to use the Data object.

apuignav:
I think we need two functions:
  - `sample`, which immediatly returns a dataset
  - `get_sampler`, which can be used as a placeholder to create loss functions and so on, and with a simple `sampler.sample` function we would get the same thing as with `pdf.sample`

mayou36:
Hm, I am not sure about the differences. My idea:

`pdf.sample` returns a `SampleData` (subclass of `Data`) containing the sample (and can be used directly with toys, anything, loss whatever)

`SampleData` holds an additional method `SampleData.resample()` which "updateds" (newly samples) the data. This applies for all the places where the sample was used before (so in the loss etc.)

This should cover everything at once. Or do you have an uncovered/dangerous usecase? Or just a suspicion? (TF is always tricky...)

apuignav:
Yeah, this works too, but you've run TF once more than you need, right?

Issue #67, opened 2018-11-23
Extended Sampling


mayou36:
Add extended sampling correctly dealing with the yield

PR #96, opened 2019-02-05
Sample data

mayou36:
solves #83 #67

### DIFF
--- a/tests/test_extended.py
+++ b/tests/test_extended.py
@@ -17,7 +17,7 @@ def test_extract_extended_pdfs():
     gauss5 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
     gauss6 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
 
-    yield1 = zfit.Parameter('yield1', 200.)
+    yield1 = zfit.Parameter('yield123' + str(np.random.random()), 200.)
 
     # sum1 = 0.3 * gauss1 + gauss2
     gauss3_ext = 45. * gauss3
@@ -36,4 +36,7 @@ def test_extract_extended_pdfs():
     limits = limits.with_autofill_axes()
     extended_sample = extended_sampling(pdfs=sum_all, sampling_func=sum_all._hook_sample, limits=limits)
     extended_sample_np = zfit.run(extended_sample)
-    assert np.shape(extended_sample_np)[0] == pytest.approx(expected=(45 + 100 + 200), rel=0.2)
+    assert np.shape(extended_sample_np)[0] == pytest.approx(expected=(45 + 100 + 200), rel=0.1)
+    samples_from_pdf = sum_all.sample(n='extended', limits=limits)
+    samples_from_pdf_np = zfit.run(samples_from_pdf)
+    assert np.shape(samples_from_pdf_np)[0] == pytest.approx(expected=(45 + 100 + 200), rel=0.1)
--- a/tests/test_extended.py
+++ b/tests/test_extended.py
@@ -8,7 +8,7 @@ from zfit.core.sample import extract_extended_pdfs, extended_sampling
 obs1 = 'obs1'
 
 
-@pytest.mark.flaky(reruns=5)
+@pytest.mark.flaky(reruns=3)  # poissonian sampling
 def test_extract_extended_pdfs():
     gauss1 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
     gauss2 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
--- a/tests/test_extended.py
+++ b/tests/test_extended.py
@@ -1,0 +1,39 @@
+import pytest
+import numpy as np
+import tensorflow as tf
+
+import zfit
+from zfit.core.sample import extract_extended_pdfs, extended_sampling
+
+obs1 = 'obs1'
+
+
+@pytest.mark.flaky(reruns=5)
+def test_extract_extended_pdfs():
+    gauss1 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
+    gauss2 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
+    gauss3 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
+    gauss4 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
+    gauss5 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
+    gauss6 = zfit.pdf.Gauss(obs=obs1, mu=1.3, sigma=5.4)
+
+    yield1 = zfit.Parameter('yield1', 200.)
+
+    # sum1 = 0.3 * gauss1 + gauss2
+    gauss3_ext = 45. * gauss3
+    gauss4_ext = 100. * gauss4
+    sum2_ext_daughters = gauss3_ext + gauss4_ext
+    sum3 = 0.4 * gauss5 + gauss6
+    sum3_ext = sum3.create_extended(yield1)
+
+    sum_all = zfit.pdf.SumPDF(pdfs=[sum2_ext_daughters, sum3_ext])
+    sum_all.set_norm_range((-5, 5))
+
+    extracted_pdfs = extract_extended_pdfs(pdfs=sum_all)
+    assert frozenset(extracted_pdfs) == {gauss3_ext, gauss4_ext, sum3_ext}
+
+    limits = zfit.Space(obs=obs1, limits=(-4, 5))
+    limits = limits.with_autofill_axes()
+    extended_sample = extended_sampling(pdfs=sum_all, sampling_func=sum_all._hook_sample, limits=limits)
+    extended_sample_np = zfit.run(extended_sample)
+    assert np.shape(extended_sample_np)[0] == pytest.approx(expected=(45 + 100 + 200), rel=0.2)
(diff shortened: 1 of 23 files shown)


zfit/core/basemodel.py
```diff
@@ -755,8 +755,18 @@ def sample(self, n: int, limits: ztyping.LimitsType, name: str = "sample") -> zt
         Returns:
             Tensor(n_obs, n_samples)
         """
+        if fixed_params is True:
```
apuignav: I don't like `fixed_params`, maybe better `freeze_params`. @rsilvaco?

zfit/core/basemodel.py
```diff
@@ -755,8 +755,18 @@ def sample(self, n: int, limits: ztyping.LimitsType, name: str = "sample") -> zt
...
+            fixed_params = list(self.get_dependents(only_floating=False))
+        if fixed_params is False:
```
apuignav: why not `else`?

mayou36: True

zfit/core/basemodel.py
mayou36: There's more to that: what about randomization? Shouldn't the data have something like: Data.sample(randomize={True, list_with_params})

apuignav:
I am still not sure about doing an unnecessary sampling.

Why don't we have a `get_sampler` that returns `SampleData`, which then you can use to `sample`.

If you want a single sample, you can do `pdf.get_sampler().sample(n=N)`. I am not sure, what do you think? @rsilvaco?

zfit/core/basemodel.py
apuignav: I'm not sure I follow, can you give me an example of `randomize`?

rsilvaco: not sure if I understood this point either. You want to keep some parameter completely fixed in the sampling?
For @apuignav, I don't have a strong preference on the name. Why you have a preference for `freeze_params`?

rsilvaco:
> If you want a single sample, you can do `pdf.get_sampler().sample(n=N)`. I am not sure, what do you think? @rsilvaco?

@apuignav: I don't like the unnecessary sampling either, but I cannot see how much better is your suggestion. Because you would in any case have this single sample reuired, right? Or am I missing the point in here?

apuignav:
No, in my suggestion we would not generate anything, because we would never run the graph, right?

rsilvaco:
> No, in my suggestion we would not generate anything, because we would never run the graph, right?

I see, then I would agree with that - though I think at some point we could think in something different ...

apuignav:
What about a flag in sample, something like `run=False`? Or simply generate 0 events and that's it?

apuignav:
`create_sample` or `create_sampler`?

mayou36:
> `create_sample` or `create_sampler`?

Typo, it's `create_sampler`.

I found it slightly better then `get_sampler`, but no strong opinion.

apuignav:
Agreed. I like `create_sampler`.

apuignav:
Should `resample` have something like `poisson` option? same as `sample`...

mayou36:
We have a few options:
What I've added now is that `n` can be a Tensor, which means it could be tf.random_poisson(...) (in `create_sampler` or `sample`) and therefore each time `resample` is called, a new sized sample (sample size drawn newly from the poissonian) is generated. This can of course be also a fixed number as up to now or an arbitrary complicated tensor (it's a little bit like with constraints: they can be arbitrary Tensors, but for the often use case we should offer a simple way).
This could be a "poisson" or a function like `zfit.sample.poisson` (which may contains options for more complicated, combined or similar). So it would be: `model.create_sampler(n=zfit.sample.poisson(extended_pdf=model), limits=...)`
(but too cumbersome?)

An other thing added:
in `Sampler.resample(fixed_params, n=None)` allows to overwrite n for this one draw with a fixed number.

comments, @apuignav

apuignav:
OK, let's go part by part

> What I've added now is that `n` can be a Tensor, which means it could be tf.random_poisson(...) (in `create_sampler` or `sample`) and therefore each time `resample` is called, a new sized sample (sample size drawn newly from the poissonian) is generated. This can of course be also a fixed number as up to now or an arbitrary complicated tensor (it's a little bit like with constraints: they can be arbitrary Tensors, but for the often use case we should offer a simple way).

That is very elegant, but maybe a bit error prone. However, as an "advanced" feature I really like it.

> This could be a "poisson" or a function like `zfit.sample.poisson` (which may contains options for more complicated, combined or similar). So it would be: `model.create_sampler(n=zfit.sample.poisson(extended_pdf=model), limits=...)`
(but too cumbersome?)

This depends on how `zfit.sample.poisson`, as we need to make sure that extended models (made of sum of extended) are fluctuated correctly. @rsilvaco did we ever solve how this had to be done?

On another point, it's true that the notation is slightly cumbersome, but on the other hand it's very explicit, so I like it. Not sure, let's see what @rsilvaco and the beta testers think?

> An other thing added:
in `Sampler.resample(fixed_params, n=None)` allows to overwrite n for this one draw with a fixed number.

That is a good idea.

mayou36:
Yes, we have to make sure that that's done right. All the logic goes into `zfit.sample.poisson`.

I'll add it, let's see what people say

apuignav:
Can we maybe do `model.create_sampler(n=N, extended=True)`? It's less correct but a bit more clear? @rsilvaco?

mayou36:
But with n specified? shouldn't the yield be taken from the model directly then? Just an "extended" option that reads as "use yield to draw n poissonian"? `Extended`, why not? Or even `n="extended"`?

I am rather confused what specifying a different n than the yield _and_ extended=True should give?

apuignav:
That's a good point. So we would have a simple case

`pdf.create_sampler(n=1000, limits=...)`

a more advanced case

`pdf.create_sampler(n=ztf.random_poisson(1000), limits=...)`

then the full general extended

`pdf.create_sampler(n=zfit.sample.poisson(extended_pdf=pdf), limits=...)`

which is annoying because you use pdf twice. And then you propose something like

`pdf.create_sampler(n="extended", limits=...)`

?

Or what about

`pdf.create_extended_sampler(limits=...)`?

(I guess limits are optional everywhere)

mayou36:
> `pdf.create_sampler(n=1000, limits=...)`
> 
> a more advanced case
actually the most general case
> 
> `pdf.create_sampler(n=ztf.random_poisson(1000), limits=...)`
and as a detail: they are basically the same, you enter both times a "Tensor-like" object :)

> then the full general extended
> 
> `pdf.create_sampler(n=zfit.sample.poisson(extended_pdf=pdf), limits=...)`
> 
> which is annoying because you use pdf twice. 
agree, and we can get rid of it:

> And then you propose something like
> `pdf.create_sampler(n="extended", limits=...)`
yes, instead of `zfit.sample.poisson`.

> Or what about
> 
> `pdf.create_extended_sampler(limits=...)`?
> 
> (I guess limits are optional everywhere)
I don't like it, just a new name for something minor.

The string works nicely if we only have a few string options (probably also "extended_blabla" if there is one other obvious way to do it). 

So my idea: n is either a (fixed) Tensor-like object or a string.
One idea I had: if n is None and the model is extended, we could automatically sample extended. Too implicit?

apuignav:
TBH, I'm not sure it's something minor, but OK...

Regarding the `n=None` (or nothing), I like it, but I'm not sure what would be the default behaviour. I guess we could do:
  - If it's extended:
    + If `n` is not given, we do extended generation.
    + If `n` is a tensor, we do whatever the tensor does.
    + If `n` is a string, we can have "poisson" or "fixed".
 - If it's not extended, `n` needs to be given, and we can basically only accept tensors.

@rsilvaco ?

mayou36:
> TBH, I'm not sure it's something minor, but OK...

Ah no, but in the end it's all about one argument, `n`, which will be different, right? Because even if we have `create_extended_sampler`, it is still not clear how the extended should exactly be done, so we need additional arguments anyway.

Or let's say: can we put everything needed to determine how to do extended/non-extended sampling into `n` (my implication was yes) or is it too unambiguous (then a second method could make sense)?

> Regarding the `n=None` (or nothing), I like it, but I'm not sure what would be the default behaviour. I guess we could do:
> 
> * If it's extended:
>   
>   * If `n` is not given, we do extended generation.
>   * If `n` is a tensor, we do whatever the tensor does.
>   * If `n` is a string, we can have "poisson" or "fixed".

Yes, agree

> * If it's not extended, `n` needs to be given, and we can basically only accept tensors.

Exactly, agree

tests/test_extended.py
```diff
@@ -0,0 +1,39 @@
...
+import numpy as np
+import tensorflow as tf
+
+import zfit
+from zfit.core.sample import extract_extended_pdfs, extended_sampling
+
+obs1 = 'obs1'
+
+
+@pytest.mark.flaky(reruns=5)
```
apuignav: Why flaky?

zfit/core/basemodel.py
```diff
@@ -742,8 +741,26 @@ def _inverse_analytic_integrate(self, x):
         else:
             return self._inverse_analytic_integral[0](x=x, params=self.params)
 
-    def create_sampler(self, n: ztyping.nSamplingTypeIn, limits: ztyping.LimitsType, fixed_params=True,
-                       name: str = "sample") -> ztyping.XType:
+    def create_sampler(self, n: ztyping.nSamplingTypeIn, limits: ztyping.LimitsType,
+                       fixed_params: Union[bool, List[ZfitParameter], Tuple[ZfitParameter]] = True,
+                       name: str = "sampler") -> "Sampler":
+        """Create a `Sampler` that acts as `Data` but can be resampled, also with changed parameter, n.
+
```
apuignav: Remove extra whitespace.

tests/test_extended.py
mayou36: Because of the sampling, how should we verify that it sampled poissonian? but 3 reruns is ok as well.

It's hard in the end to test if the sampling is correct, we just check the sum to be roughly the same now

zfit/core/basemodel.py
mayou36: done

mayou36:
Looks good so far, what do you think, @apuignav ?

PR #97, opened 2019-02-05
Simple subclass

mayou36:


apuignav:
I like where this is going :-)

zfit/core/basemodel.py
```diff
@@ -900,3 +901,47 @@ def __mul__(self, other):
...
+
+    Example:
+
+    .. code:: python
+
+        class MyPDF(zfit.pdf.ZPDF):
+            _PARAMS = ['mu', 'sigma']
+
+            def _unnormalized_pdf(self, x):
+                # mu = self.params['mu']
```
apuignav: Should these be uncommented?

mayou36: Yes, good catch!

mayou36:
Ready from my side

apuignav:
LGTM!

PR #99, opened 2019-02-13
Fix exponential pdf

mayou36:
Introducing a shift in exp, leaving the pdf untouched and shifting x to around 0 for numerical stability (exp overflows otherwise)

Also changed analytic integral function API from
`func(limits, params)` 
to
`func(limits, params, model)` 
-> some code may breaks

### DIFF
--- a/tests/test_basePDF.py
+++ b/tests/test_basePDF.py
@@ -126,7 +126,7 @@ def test_analytic_sampling():
     class SampleGauss(TestGaussian):
         pass
 
-    SampleGauss.register_analytic_integral(func=lambda limits, params: 2 * limits.upper[0][0],
+    SampleGauss.register_analytic_integral(func=lambda limits, params, model: 2 * limits.upper[0][0],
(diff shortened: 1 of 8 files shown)


apuignav:
What's the use of `model`?

mayou36:
So this is the model that get's integrated.

For generality, mostly. In case _any_ information/method from the model is needed, (like things not contained in params, as for example daughter pdfs) it can be accessed like this. This allows for the max freedom (and rather introduce it now then braking it later).

Current explicit usecase: a custom method in the model and a "shift state".

apuignav:
I am not sure I understand. When you register you have access to the model anyway right?

mayou36:
Not to the instance, no. Only to the class. So if you have an integral that depends on something instance specific that is not in `params`, then you need `model`.

(sorry for merging already, thought was cleared)

apuignav:
No problem! I guess I'll understand it when I see it in use

PR #102, opened 2019-02-14
Simpler limits

mayou36:
currently `limit1d` to return (lower, upper) for simple scalars
what other shortcut would be helpful, 1d with two limits? @apuignav


apuignav:
We discussed three things:
  - 1D limit. 
  - disjoint 1D limit.
  - Square 2D limit. 

This implements first one, so I would add the other two.

apuignav:
Wonderful, merging

PR #106, opened 2019-02-15
Expand capabilities of dealing with complex parameters

apuignav:
- [x] Implement `PolarComplexParameter`, working with tensors.
- [x] Add `conj` method to `*ComplexParameter`s.
- [x] Tests

### DIFF
--- a/zfit/core/parameter.py
+++ b/zfit/core/parameter.py
@@ -464,42 +464,51 @@ class BaseComposedParameter(ZfitParameterMixin, ComposedVariable, BaseParameter)
 class ComposedParameter(BaseComposedParameter):
     def __init__(self, name, tensor, **kwargs):
         tensor = ztf.convert_to_tensor(tensor)
-        independend_params = tf.get_collection("zfit_independent")
-        params = get_dependents(tensor=tensor, candidates=independend_params)
+        independent_params = tf.get_collection("zfit_independent")
+        params = get_dependents(tensor=tensor, candidates=independent_params)
         # params_init_op = [param.initializer for param in params]
         params = {p.name: p for p in params}
         # with tf.control_dependencies(params_init_op):
         super().__init__(params=params, initial_value=tensor, name=name, **kwargs)
 
 
-class BaseComplexParameter(BaseComposedParameter):
-    def __init__(self, name, initial_value, floating=True, dtype=ztypes.complex, **kwargs):
-        initial_value = tf.cast(initial_value, dtype=dtype)
-        real_value = tf.real(initial_value)
-        real_part = Parameter(name=name + "_real", init_value=real_value, floating=floating, dtype=real_value.dtype)
-        imag_value = tf.imag(initial_value)
-        imag_part = Parameter(name=name + "_imag", init_value=imag_value, floating=floating, dtype=imag_value.dtype)
-        params = {'real': real_part, 'imag': imag_part}
-        super().__init__(params=params, initial_value=initial_value, name=name, **kwargs)
+class ComplexParameter(BaseComposedParameter):
+    def __init__(self, name, initial_value, dtype=ztypes.complex, **kwargs):
+        tensor = ztf.convert_to_tensor(initial_value, dtype=dtype)
+        independent_params = tf.get_collection("zfit_independent")
+        params = get_dependents(tensor=tensor, candidates=independent_params)
+        params = {p.name: p for p in params}
+        super().__init__(name=name, params=params, initial_value=tensor, **kwargs)
+
+    @staticmethod
+    def from_cartesian(name, real, imag, dtype=ztypes.complex, **kwargs):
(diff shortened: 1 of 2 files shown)


apuignav:
So far, I think this only works with floats as inputs, no tensors, but I need @mayou36 help for that.

mayou36:
We could even try to make it better: the only difference between the angular and the complex is the constructor, right? So we can move everything from ComplexParamtere to BaseComplexParameter and inherit from that (so that `ComplexParameter` is basically an "empty" inherit, `ComplexPolarParameter` adds somethings in the init.

apuignav:
Done! What do we gain with this, though?

mayou36:
Hm, no, I have to think it through. Feels like somethings still odd

apuignav:
OK, think about it, I'll write a couple of tests in the meantime.

zfit/core/parameter.py
```diff
@@ -465,23 +465,51 @@ class ComposedParameter(BaseComposedParameter):
...
+
+    @staticmethod
+    def from_polar(name, abs, arg, dtype=ztypes.complex, **kwargs):
+        return ComplexParameter(name, tf.cast(tf.complex(abs*tf.math.cos(arg),
+                                                         abs*tf.math.sin(arg)),
+                                              dtype=dtype))
+
+    def conj(self):
+        return ComposedParameter('{}_conj'.format(self.name),
+                                 tf.math.conj(self))
```
mayou36: This should be a ComplexParameter as well, right?

And all the arguments should be propagated, e.g. floating=self.floating, dtype=...

zfit/core/parameter.py
```diff
@@ -465,23 +465,51 @@ class ComposedParameter(BaseComposedParameter):
...
-        params = get_dependents(tensor=tensor, candidates=independend_params)
+        independent_params = tf.get_collection("zfit_independent")
+        params = get_dependents(tensor=tensor, candidates=independent_params)
         # params_init_op = [param.initializer for param in params]
         params = {p.name: p for p in params}
         # with tf.control_dependencies(params_init_op):
         super().__init__(params=params, initial_value=tensor, name=name, **kwargs)
 
 
 class ComplexParameter(BaseComposedParameter):
```
mayou36: Actually, why not just inherit from ComposedParameter? It *is* a composed parameter and reduces code dublication

zfit/core/parameter.py
```diff
@@ -465,23 +465,51 @@ class ComposedParameter(BaseComposedParameter):
...
+        params = get_dependents(tensor=tensor, candidates=independent_params)
+        params = {p.name: p for p in params}
+        super().__init__(name=name, params=params, initial_value=tensor, **kwargs)
+
+    @staticmethod
+    def from_cartesian(name, real, imag, dtype=ztypes.complex, **kwargs):
+        return ComplexParameter(name, tf.cast(tf.complex(real, imag), dtype=dtype))
+
+    @staticmethod
+    def from_polar(name, abs, arg, dtype=ztypes.complex, **kwargs):
```
mayou36: abs -> mod?

so we have name consistency with the attribute and no clash with builtins

zfit/core/parameter.py
```diff
@@ -465,23 +465,51 @@ class ComposedParameter(BaseComposedParameter):
...
+                                              dtype=dtype))
+
+    def conj(self):
+        return ComposedParameter('{}_conj'.format(self.name),
+                                 tf.math.conj(self))
+
+    @property
+    def real(self):
+        return tf.real(self.value())
+
```
mayou36: Just `self` is enough, remove the `.value()`. `value()` does a certain read, there would also be `read_value`. It's kinda unclear tough, but to be safe (and something for the new tf Variables to see how they do it/recommend ;) )

zfit/core/parameter.py
apuignav: Ah yes, `floating` and `dtype`?

zfit/core/parameter.py
apuignav: Oh, so true! Fixed!

zfit/core/parameter.py
mayou36: Yes, that's it

zfit/core/parameter.py
apuignav: OK!

zfit/core/parameter.py
apuignav: OK!

zfit/core/parameter.py
apuignav: Since `convert_to_tensor` has a hardcoded `dtype.float`, I updated `ComposedParameter`.

mayou36:
Nice catch with the `ztf.convert_to_tensor` dtype!

I'll gonna look it through tomorrow, just to think through again but looks good (I can also resolve the conflicts)

apuignav:
OK, I let you to fix conflicts and merge ;-)

apuignav:
Thanks! That was really useful!

PR #107, opened 2019-02-17
Improve caching

mayou36:


apuignav:
That's interesting, can you add some docs on how the caching method works, so advanced users can take advantage of it?

mayou36:
Sure, I've added a few lines at the module level. It's quite primitive caching and it can be improved of course, but let's see what the needs are.

I've set python 3.5 to allow for failure, do you think we can drop it? And only support 3.6, 3.7? With 3.5 we have a metaclass conflict that is basically unsolvable (it needs two metaclasses currently: tf.Varialbe metaclass as we inherit and another metaclass that backports a feature, the `__init_subclass__`, which is essential).

If we create our own Variables, we can maybe bring it back. But is 3.5 critical? I see a lot of packages moving already/soon away from it, seems like the `3.*` have shorter lifetimes.

apuignav:
Fine by me, let's drop 3.5. I'll check the docs and approve the PR this morning.

PR #109, opened 2019-02-18
Fix consistency of init_value.

### LINKED ISSUE #108: Consistency in kwarg naming in parameters
In some cases, we use `init_value` and in some others `initial_value`. Should we make this consistent?

apuignav:
Closes #108.

PS: I only fixed the "user-facing" kwargs, that is, from `Parameter` onwards.


mayou36:
Hm, thinking about it: is just `value` a better name? The `init value` comes more from the TF side but is actually "hidden" and rather confusing thinking about it.

just "value"?

apuignav:
Why not? Let me replace all `init_value` by `value`.

PR #110, opened 2019-02-21
Pandas data

mayou36:
Add `from_pandas` and `to_pandas` function. This allows for simple preprocessing: load `from_root`, do `to_pandas`, preprocess, `from_pandas`

Sure, more overhead than doing it manually with uproot + pandas, but it works


mayou36:
Yes, most of all because it's a one-time thing anyway

PR #126, opened 2019-03-26
add double_crystalball_func to zfit/models/physics

marinang:
@apuignav @mayou36   
Do you want me to also create the PDF class for the `DoubleCrystallBall`?

### DIFF
--- a/tests/test_pdf_cb.py
+++ b/tests/test_pdf_cb.py
@@ -31,6 +31,7 @@ def eval_testing(pdf, x):
 
 
 def test_cb_dcb():
+    zfit.run.create_session(reset_graph=True)
 
     obs = zfit.Space('x', limits=bounds)
 
--- a/tests/test_pdf_cb.py
+++ b/tests/test_pdf_cb.py
@@ -1,0 +1,94 @@
+import pytest
+import numpy as np
+from zfit.models.physics import CrystalBall, DoubleCB
+import zfit
+from scipy.stats import crystalball
+
+mu = 0.0
+sigma = 0.5
+
+alphal = 1.0
+nl = 2.0
+
+alphar = 3.0
+nr = 4.0
+
+bounds = (-10, 3.0)
+lbounds = (bounds[0], mu)
+rbounds = (mu, bounds[1])
+
+
+def sample_testing(pdf):
+    sample = pdf.sample(n=1000)
+    sample_np = zfit.run(sample)
+    assert not any(np.isnan(sample_np))
+
+
+def eval_testing(pdf, x):
+    probs = zfit.run(pdf.pdf(x))
+    assert not any(np.isnan(probs))
+    return probs
+
+
+def test_cb_dcb():
+
+    obs = zfit.Space('x', limits=bounds)
+
+    mu_ = zfit.Parameter('mu_cb', mu)
+    sigma_ = zfit.Parameter('sigma_cb', sigma)
+    alphal_ = zfit.Parameter('alphal_cb', alphal)
+    nl_ = zfit.Parameter('nl_cb', nl)
+    alphar_ = zfit.Parameter('alphar_cb', alphar)
+    nr_ = zfit.Parameter('nr_cb', nr)
+
+    cbl = CrystalBall(obs=obs, mu=mu_, sigma=sigma_, alpha=alphal_, n=nl_)
+    cbr = CrystalBall(obs=obs, mu=mu_, sigma=sigma_, alpha=-alphar_, n=nr_)
+    dcb = DoubleCB(obs=obs, mu=mu_, sigma=sigma_, alphal=alphal_, nl=nl_, alphar=alphar_, nr=nr_)
+
+    sample_testing(cbl)
+    sample_testing(cbr)
+    sample_testing(dcb)
+
+    x = np.random.normal(mu, sigma, size=10000)
+
+    probsl = eval_testing(cbl, x)
+    probsr = eval_testing(cbr, x)
+
+    assert not any(np.isnan(probsl))
+    assert not any(np.isnan(probsr))
+
+    probsl_scipy = crystalball.pdf(x, beta=alphal, m=nl, loc=mu, scale=sigma)
+    probsr_scipy = crystalball.pdf(-x, beta=alphar, m=nr, loc=mu, scale=sigma)
+
+    ratio_l = probsl_scipy / probsl
+    ratio_r = probsr_scipy / probsr
+
+    assert np.allclose(ratio_l, ratio_l[0])
+    assert np.allclose(ratio_r, ratio_r[0])
+
+    kwargs = dict(limits=(-5.0, mu), norm_range=lbounds)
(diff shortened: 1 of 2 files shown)


mayou36:
Yes please, go ahead! While the analytic integral has some problems, a bug with tf, can you add this as well? The integral for a simple CB is there

marinang:
@mayou36 you mean create an integral function from the already existing `crystalball_integral` function ?

mayou36:
Yes, creating the integral from scratch is slightly cumbersome (as can be seen in the `crystalball_integral`) but using the existing function (doing basically the same as for the function) it should be easy (with a `cond` I think). Let me know if you prefer me to do it, but I'd suggest just try and then we discuss, ok?

Btw: can you add a test to it and docs? For the test, we did not yet fully formalize the tests, so may look at an existing test of e.g. exponential, gaussian,  ok?

marinang:
@mayou36  Yep I will do that.

marinang:
Tests are on their way. 

I thought `DoubleCB`  was less heavy than `DoubleCrystallBall`, but I can change if you prefer the latter.

mayou36:
Looks good so far!

Put it, for the time being, in test_pdf_doublecb.py. We're currently refactoring the tests (#131), make sure that any graph is only built inside the test function(s) and that you clean the graph in the beginning by putting  `zfit.run.create_session(reset_graph=True)` at the top of the function call.

Probably wait with the tests a few days, we can discuss on #131 what to automatically test etc, so that you don't have to double the work. Some tests will then be done automatically

marinang:
@mayou36 alright I will wait.

marinang:
@mayou36 #131 was merged, so I can move on with the test?

mayou36:
Yes, but it's not the final testing. Implement tests **except** of
* testing the normalization
* comparing the analytic integral with the numeric

But otherwise add any test that you think fits, e.g. cross comparison with CB and SumPDF? This should give the same basically, right? (if you choose the right shared Parameters)

Can you add a test to compare it with scipy dist probably? I was thinking of adding a general comparison vs scipy dist, so if you come up with a clean test I extract a more general case.

marinang:
@mayou36 Good idea, and especially to compare with scipy is quite robust I think.

apuignav:
@marinang Sorry for all this requests, we're still figuring out what's the needed testing for PDFs and things like that. I would second @mayou36's comment: either provide a small file (ROOT, hdf) with the behaviour you want, or use a different generator like `scipy.dist`.

I would suggest comparing histograms with `scipy.stats.ks_2amp`, similarly to what is done in https://github.com/zfit/phasespace/blob/master/tests/test_physics.py#L95. What do you think?

marinang:
@apuignav do you mean create one sample from the DCB/CB from zfit and one from Scipy and do a KS  test ? This would be seed dependents in a test.

apuignav:
I was thinking something like that. If we have enough events, it should not be too dependent.

marinang:
@apuignav ok but maybe we should not be too strict with the KS test, anyway I will try and see.

mayou36:
sampling is one thing, the more important thing I think is to compare the pdf values. So for certain x given to scipy dist, it should return the same for zfit (with a different normalization, as scipy dist is normalized over -inf, inf). Dividing the two values should return an array filled with basically one number (the ratio).

The normalization of the zfit pdf is tested somewhere else, so the ratio above can be neglected, it just has to be the same over the whole array.
(Or we even use the scipy integral to remove the factor, even cleaner.)

On sampling, we can also mark the test flaky, so it repeats a few times if failure, rendering the possibility of test failure to nearly 0
@marinang I would say put there what you think is reasonable, a pdf comparison and a sampling seems to me the way to go, but as mentioned, we're also still figuring out what's right

apuignav:
You are right, it's a PDF. We should generate random x values and compare. KS doesn't make sense here.

mayou36:
Looks good so far, that's the direction I was thinking of. Let me know once it's not WIP anymore and I'll review it fully

marinang:
@mayou36 I just removed the WIP, you can review it.

tests/test_pdf_cb.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def eval_testing(pdf, x):
+    probs = zfit.run(pdf.pdf(x))
+    assert not any(np.isnan(probs))
+    return probs
+
+
+def test_cb_dcb():
+
```
mayou36: Add here the line
`zfit.run.create_session(reset_graph=True)`

this resets the graph and session to make sure the test is independent

marinang:
@mayou36 test fails if I add `zfit.run.create_session(reset_graph=True)`

```
===================================================================================================== FAILURES ======================================================================================================
____________________________________________________________________________________________________ test_cb_dcb ____________________________________________________________________________________________________

    def test_cb_dcb():
>       zfit.run.create_session(reset_graph=True)

test_pdf_cb.py:34:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <zfit.util.execution.RunManager object at 0x111ee5940>, args = (), kwargs = {'reset_graph': True}, sess_kwargs = {'reset_graph': True}

    def create_session(self, *args, **kwargs):
        """Create a new session (or replace the current one). Arguments will overwrite the already set arguments.

        Args:
            *args ():
            **kwargs ():

        Returns:
            :py:class:`tf.Session`
        """
        sess_kwargs = copy.deepcopy(self._sess_kwargs)
        sess_kwargs.update(kwargs)
>       self.sess = tf.Session(*args, **sess_kwargs)
E       TypeError: __init__() got an unexpected keyword argument 'reset_graph'

../zfit/util/execution.py:91: TypeError
----------------------------------------------------------------------------------------------- Captured stderr call ------------------------------------------------------------------------------------------------
Exception ignored in: <bound method BaseSession.__del__ of <tensorflow.python.client.session.Session object at 0x1a361b58d0>>
Traceback (most recent call last):
  File "/Users/matthieumarinangeli/anaconda3/envs/tfn/lib/python3.6/site-packages/tensorflow/python/client/session.py", line 736, in __del__
    if self._session is not None:
AttributeError: 'Session' object has no attribute '_session'
```

mayou36:
Yes, exactly, sorry. Can you get the changes in (e.g. rebase the branch on develop and force-push)?

This makes sure that the graph is clean

marinang:
@mayou36 I know the test is working so I will just push and see here if anything complains, there shouldn't be any conflict.

mayou36:
Yes of course, the PR CI should pass

PR #128, opened 2019-03-27
Fix typo in documentation

chrisburr:


apuignav:
Good catch, thanks a lot!

PR #129, opened 2019-03-28
Fix sampling

mayou36:


apuignav:
Should we make a list of things to fix with TF2.0?

mayou36:
I think there are anyway so many things, it's not worth. And anything tf.Variable related will be looked at

PR #132, opened 2019-04-02
Small update in CONTRIBUTING guidelines

apuignav:
Very small change, but we can use this as guideline for `zfit-physics`. Anything else that should be added in the guidelines? Coding standards maybe?

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,16 +1,27 @@
-* zfit version:
-* Python version:
-* Operating System:
-* Tensorflow version:
+<!--- Provide a general summary of the issue in the Title above -->
 
-### Description
+## Current Behaviour
 
-Describe what you were trying to get done.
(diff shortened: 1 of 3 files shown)


mayou36:
I think we should try to make the file shorter and put some information to other places. For example to remove the how-to in the bug report and instead use only the issue-template. On the other hand, we could move out some of the explicit guidelines to the Wiki of the repo (like "cloning the repo" etc), anything that is not _explicitly_ related to zfit contributions in the coding sense. That's the direction I see, what do you think?

One explicit thing to add soon would be "how to contribute a pdf" for example, explaining which unittests etc are needed.

apuignav:
I've added a couple of simple templates. Regarding the "how to contribute a PDF" and so on, I think this belongs to `zfit-physics`. What do you think?

mayou36:
Yes, this looks good to me. And yes, this belongs there. We can add a note here in contrib once we have it in zfit-physics to point there

PR #147, opened 2019-04-17
Tfp dists

mayou36:
Changes:
- Move the instantiation of the tfp distributions inside the `unnormalized_pdf`, s.t. the parameters op connecting it with the output happen inside it. Needed for the gradient taping (https://github.com/tensorflow/probability/issues/285) as in normalization on-the-fly

- add the truncatedgauss (simple TFP wrap)
- add docs to Gauss, Uniform
- use setup_function and teardown_function in every unittest module (currently resets tf session and graph)


apuignav:
Hey, can you add a description of this PR for posterity?

mayou36:
yep, of course

PR #150, opened 2019-04-29
Polynomials from recurrence relations.

apuignav:
Pending:
  - [x] Analytic integrals
  - [x] Tests

### DIFF
--- a/tests/test_pdf_polynomials.py
+++ b/tests/test_pdf_polynomials.py
@@ -25,12 +25,13 @@ coeffs_parametrization = [
 ]
 
 rel_integral = 6e-2
+default_sampling = 60000
 
-poly_pdfs = [zfit.pdf.Legendre,
-             zfit.pdf.Chebyshev,
-             zfit.pdf.Chebyshev2,
-             zfit.pdf.Hermite,
-             zfit.pdf.Laguerre]
+poly_pdfs = [(zfit.pdf.Legendre, default_sampling),
+             (zfit.pdf.Chebyshev, default_sampling),
+             (zfit.pdf.Chebyshev2, default_sampling),
+             (zfit.pdf.Hermite, default_sampling * 20),
+             (zfit.pdf.Laguerre, default_sampling)]
 
 
 @pytest.mark.parametrize("coeffs", coeffs_parametrization)
@@ -50,10 +51,11 @@ def test_legendre_polynomial(coeffs):
     # assert pytest.approx(zfit.run(numerical_integral2), rel=1e-2) == zfit.run(integral2)
 
 
-@pytest.mark.parametrize("poly_pdf", poly_pdfs)
+@pytest.mark.parametrize("poly_cfg", poly_pdfs)
 @pytest.mark.parametrize("coeffs", coeffs_parametrization)
-def test_polynomials(poly_pdf, coeffs):
+def test_polynomials(poly_cfg, coeffs):
     coeffs = copy.copy(coeffs)
+    poly_pdf, n_sampling = poly_cfg
     polynomial = poly_pdf(obs=obs1, coeffs=coeffs)
 
     # test 1 to 1 range
@@ -79,6 +81,6 @@ def test_polynomials(poly_pdf, coeffs):
     assert pytest.approx(analytic_integral, rel=rel_integral) == zfit.run(numerical_integral)
 
     lower, upper = obs1_random.limit1d
-    test_integral = np.average(zfit.run(polynomial.unnormalized_pdf(tf.random.uniform((100000,), lower, upper)))) \
+    test_integral = np.average(zfit.run(polynomial.unnormalized_pdf(tf.random.uniform((n_sampling,), lower, upper)))) \
                     * obs1_random.area()
     assert pytest.approx(analytic_integral, rel=rel_integral) == test_integral
--- a/tests/test_pdf_polynomials.py
+++ b/tests/test_pdf_polynomials.py
@@ -13,16 +13,14 @@ obs1 = zfit.Space(obs="obs1", limits=(-0.9, 0.85))
 def test_legendre_polynomial():
     legendre = zfit.pdf.Legendre(obs=obs1, coeffs=[0.1, 1.5, 0.6])
 
-    integral = legendre.integrate(limits=obs1)
-    assert pytest.approx(1, rel=1e-3) == zfit.run(integral)
-    numerical_integral = legendre.numeric_integrate(limits=obs1)
-    assert pytest.approx(1, rel=1e-3) == zfit.run(numerical_integral)
+    integral = legendre.integrate(limits=obs1, norm_range=False)
+    numerical_integral = legendre.numeric_integrate(limits=obs1, norm_range=False)
+    assert zfit.run(numerical_integral) == pytest.approx(zfit.run(integral))
 
 
 def test_legendre_chebyshev():
     chebyshev = zfit.pdf.Chebyshev(obs=obs1, coeffs=[0.1, 0.5, 0.6, 0.3])
 
-    integral = chebyshev.integrate(limits=obs1)
-    assert pytest.approx(1, rel=1e-3) == zfit.run(integral)
-    numerical_integral = chebyshev.numeric_integrate(limits=obs1)
-    assert pytest.approx(1, rel=1e-3) == zfit.run(numerical_integral)
+    integral = chebyshev.integrate(limits=obs1, norm_range=False)
+    numerical_integral = chebyshev.numeric_integrate(limits=obs1, norm_range=False)
+    assert pytest.approx(zfit.run(integral), rel=1e-3) == zfit.run(numerical_integral)
--- a/tests/test_pdf_polynomials.py
+++ b/tests/test_pdf_polynomials.py
@@ -35,8 +35,16 @@ def test_chebyshev():
 
 
 def test_chebyshev2():
-    chebyshev2 = zfit.pdf.Chebyshev2(obs=obs1, coeffs=[1.0])
+    chebyshev2 = zfit.pdf.Chebyshev2(obs=obs1, coeffs=[2.1, 1.5, 0.6])
 
     integral = chebyshev2.analytic_integrate(limits=obs1, norm_range=False)
     numerical_integral = chebyshev2.numeric_integrate(limits=obs1, norm_range=False)
+    assert pytest.approx(zfit.run(integral), rel=1e-3) == zfit.run(numerical_integral)
+
+
+def test_laguerre():
+    laguerre = zfit.pdf.Laguerre(obs=obs1, coeffs=[1.1, 1.5, 0.6])
+
+    integral = laguerre.analytic_integrate(limits=obs1, norm_range=False)
+    numerical_integral = laguerre.numeric_integrate(limits=obs1, norm_range=False)
     assert pytest.approx(zfit.run(integral), rel=1e-3) == zfit.run(numerical_integral)
--- a/tests/test_pdf_polynomials.py
+++ b/tests/test_pdf_polynomials.py
@@ -27,7 +27,7 @@ def test_legendre_polynomial():
 
 
 def test_chebyshev():
-    chebyshev = zfit.pdf.Chebyshev(obs=obs1, coeffs=[1.1, 11.5, 0.6])
+    chebyshev = zfit.pdf.Chebyshev(obs=obs1, coeffs=[1.1, 1.5, 0.6])
 
     integral = chebyshev.integrate(limits=obs1, norm_range=False)
     numerical_integral = chebyshev.numeric_integrate(limits=obs1, norm_range=False)
--- a/tests/test_pdf_polynomials.py
+++ b/tests/test_pdf_polynomials.py
@@ -19,7 +19,7 @@ coeffs_parametrization = [
     [0.6],
     [1.42, 1.2],
     [0.2, 0.8, 0.5],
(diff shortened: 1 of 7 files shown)


mayou36:
should we make the assumption of always n and n-1 as given? (and as initial with f0, f1?). E.g. a simple polynom needs only nth step (but if it is the only exception, we may just leave it).

Detail on the order and naming: currently, f0 is the 0th element, f1 the first. But p1 is the nth element, p2 is the n-1th element, this should probably be consistent (left older, right newer element, my proposal).

apuignav:
Actually, this is just a "3-term recursive", it could be n, n-1 or something else. Although we could change it as you say, but I was trying to keep "generality".

mayou36:
Yes, true. So instead of 3 term, just n term recursive? I don't see a direct need for it TBH though, do you

apuignav:
I don't know of any polynomial that's defined with more than triple
recursion. In any case, this is an easy extension for the future.

What is missing is the fact that the observable can only be defined from -1
to 1, otherwise we need to rescale. Should it be done automatically?

Albert

mayou36:
Agree. And not automatically, we can't really know the users intention. The data should then be rescaled as well -> let's throw an error

rsilvaco:
@mayou36 start to finally checking this .... we should provide a method to automatically re-scale it to -1 to 1. This is a very specific case for the polynomial function, so we should have a flag that could enable this and therefore modify the data as well. But I do strongly prefer to have it "behind the scene".  For instance, any of these angular folded PDFs are not defined between -1 and 1, but it should be done almost by default if we want to use it

apuignav:
I agree. Data is the responsibility of the user, but scaling is ours I think.

rsilvaco:
But I must admit that I'm not sure what is the "cleanest way". For instance, assume the user defined an `obs_phi` that is between 0 and pi. This is what makes sense to the user, and then he wants to model a bkg shape using chebychev in this space. So in my mind he should be able to do `legendre = zfit.pdf.Legendre(obs=obs_phi, degree=5)` without minding the limit ... maybe some warning/some flag or something...not sure what is the simplest way ...

apuignav:
I was thinking about that. I think we need a `scale` flag (default True maybe). If true, rescales if necessary, if false either doesn't or throws an error if the limits are too large.

What do you think?

rsilvaco:
My only doubt is whether it makes sense to calculate without it. For instance, shouldn't we check if it is defined between -1 and 1 ourselves? I mean, a flag is fine, but I'm just wondering the use case of not wanting this ...

apuignav:
The use case is precisely what you said. Imagine we have (0, phi); in this case we may not want to rescale.

rsilvaco:
not sure if I understand. Let me give you a specific case: Roofit Chebychev function: https://root.cern.ch/doc/v606/RooChebychev_8cxx_source.html
In here, it does behind the scene the scaling. It is nice because I've used before for fitting the B mass without even knowing the rescaling was happening ....

apuignav:
I agree, but I would leave the possibility to turn scaling off for some special cases. In any case the default would be to rescale.

rsilvaco:
yep! fully agree!

apuignav:
I just pushed two changes:
  - Use coefficients to build the sum of polynomials.
  - Add the possibility to scale according to the observable limits.

I have not updated the tests so they will break, but it's because I don't understand them. I will add a couple of comments to the code so we can discuss them.

tests/test_pdf_polynomials.py
```diff
@@ -0,0 +1,28 @@
+#  Copyright (c) 2019 zfit
+
+
+import pytest
+import zfit
+from zfit import ztf
+
+from zfit.core.testing import setup_function, teardown_function, tester
+
+obs1 = zfit.Space(obs="obs1", limits=(-0.9, 0.85))
```
apuignav: This should be (-1, 1) to be canonical, shouldn't it?

tests/test_pdf_polynomials.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def test_legendre_polynomial():
+    legendre = zfit.pdf.Legendre(obs=obs1, degree=5)
+
+    integral = legendre.integrate(limits=obs1)
+    assert pytest.approx(1, rel=1e-3) == zfit.run(integral)
```
apuignav: Why is the integral 1 with the observable in a limited range? Am I misunderstanding something?

tests/test_pdf_polynomials.py
rsilvaco: I agree, for the canonical case this should be (-1, 1) ...
Now have you changed also this to receive any other non-canonical range and re-scale back?

apuignav: Yes, it simply scales when calculating the polynomial, building them in the interval [a,b] as orthogonal.

rsilvaco: perfect! I will give a try once I get to the office

apuignav: We need to write proper tests for it. Also, you should now be able to use the polynomials as PDFs with coefficients.

rsilvaco: ok, but what I didn't understand is wether we have already an implementation that I can provide the parameters or whether I need to build the c0*Chebby(order1) + c1*Cheby(order2) + ...?

apuignav: No, now you do

```
Cheby(obs=obs, [c0, c1, c2, ...])
```

and the degree is going to be inferred from there.

rsilvaco: perfect!

tests/test_pdf_polynomials.py
mayou36: By definition: it's the integral over the same range as the normalization range. But we don't need it here, it was more of an "make sure it runs without obvious errors" test, that the integral calculations don't break.

tests/test_pdf_polynomials.py
mayou36: It's not (-1, 1) to test the analytic integral, since (-1, 1) is already normalized to 1.
But of course, both should be tested, the tests are yet incomplete.

On the scaling: We also need to rescale the analytic integration then (the limits given there), right?
I'll do that with some clean up

apuignav: True, the analytic integral should be rescaled.

rsilvaco: Given another try on this, I see the following issue:
 `Traceback (most recent call last):
  File "simpleBkgFit.py", line 72, in <module>
    minimum = minimizer.minimize(loss=nll)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/baseminimizer.py", line 205, in minimize
    return self._hook_minimize(loss=loss, params=params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/baseminimizer.py", line 210, in _hook_minimize
    return self._call_minimize(loss=loss, params=params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/baseminimizer.py", line 214, in _call_minimize
    return self._minimize(loss=loss, params=params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/minimizer_minuit.py", line 48, in _minimize
    gradients = loss.gradients(params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 147, in gradients
    return self._gradients(params=params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 240, in _gradients
    gradients = {(p, grad) for p, grad in zip(params_todo, super()._gradients(params_todo))}
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 209, in _gradients
    return tf.gradients(self.value(), params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 187, in value
    return self._value()
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 223, in _value
    loss = super()._value()
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 192, in _value
    constraints=self.constraints)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 255, in _loss_func
    nll = _unbinned_nll_tf(model=model, data=data, fit_range=fit_range)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 38, in _unbinned_nll_tf
    for p, d, r in zip(model, data, fit_range)]
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 38, in <listcomp>
    for p, d, r in zip(model, data, fit_range)]
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/loss.py", line 42, in _unbinned_nll_tf
    probs = model.pdf(data, norm_range=fit_range)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/basepdf.py", line 314, in pdf
    value = self._single_hook_pdf(x=x, norm_range=norm_range, name=name)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/basepdf.py", line 324, in _single_hook_pdf
    return self._hook_pdf(x=x, norm_range=norm_range, name=name)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/basepdf.py", line 327, in _hook_pdf
    return self._norm_pdf(x=x, norm_range=norm_range, name=name)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/basepdf.py", line 330, in _norm_pdf
    return self._call_pdf(x=x, norm_range=norm_range, name=name)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/basepdf.py", line 338, in _call_pdf
    return self._fallback_pdf(x=x, norm_range=norm_range)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/basepdf.py", line 341, in _fallback_pdf
    pdf = self._call_unnormalized_pdf(x, name="_call_unnormalized_pdf")
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/core/basepdf.py", line 289, in _call_unnormalized_pdf
    return self._unnormalized_pdf(x)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/models/polynomials.py", line 50, in _unnormalized_pdf
    lim_low, lim_high = self.obs.limit1d
AttributeError: 'tuple' object has no attribute 'limit1d'`

apuignav: @rsilvaco Can you try again with the latest commit?

mayou36: Yes, the scaled polynomials with integral should work now (since the integral is trivially 2).

rsilvaco: Nop:
`Traceback (most recent call last):
  File "simpleBkgFit.py", line 156, in <module>
    minimum = minimizer.minimize(loss=nll)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/baseminimizer.py", line 205, in minimize
    return self._hook_minimize(loss=loss, params=params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/baseminimizer.py", line 210, in _hook_minimize
    return self._call_minimize(loss=loss, params=params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/baseminimizer.py", line 214, in _call_minimize
    return self._minimize(loss=loss, params=params)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/minimizer_minuit.py", line 50, in _minimize
    self._check_gradients(params=params, gradients=gradients)
  File "/anaconda3/envs/zfitEnvPol/lib/python3.6/site-packages/zfit/minimizers/baseminimizer.py", line 124, in _check_gradients
    "instead of a `CompositeParameter` was created implicitly.".format(non_dependents))
ValueError: Invalid gradients for the following parameters: [<tf.Variable 'a2' shape=() dtype=float64>]The function does not depend on them. Probably a Tensorinstead of a `CompositeParameter` was created implicitly.`

apuignav: Ping @mayou36?

mayou36: Ah sorry, forgot to reply: this is "fixed".

The error is that the parameter "a2" did not had a gradient, which means the pdf was independent of it (degree number wrong implemented).

@apuignav the `self.degree` = degree - 1 seems confusing to me. What't the reason for it to be there?

An the current problem: the numeric integral of the chebyshev is not working (e.g. depends on the coefficients) -> the implementation is wrong, but I do  not see what's wrong with it. If you find time (anyone) to have a look at the implementation of the chebyshev pdf, feel free. It's short but I don't find any mistake with it

mayou36:
@apuignav I'll have a take on that since its needed. I refactor to functions (so it can be lighter used to build custom pdfs etc.

The `n_degree` was changed to be n-1? Why exactly?

PR #154, opened 2019-05-16
More general nll_gaussian constraint

marinang:
nll_gaussian can now take as input for the `sigma` argument either a list of standard deviations or a covariance matrix.

## Tests added

None, already implemented tests are working.

### DIFF
--- a/tests/test_loss.py
+++ b/tests/test_loss.py
@@ -111,8 +111,8 @@ def test_unbinned_simultaneous_nll():
 
 
 @pytest.mark.flaky(3)
-@pytest.mark.parametrize('weights', (None, None, np.random.normal(loc=1., scale=0.2, size=test_values_np.shape[0]), None, None))
-@pytest.mark.parametrize('sigma', (constr, covariance, constr, constr_tf, covariance))
+@pytest.mark.parametrize('weights', (None, np.random.normal(loc=1., scale=0.2, size=test_values_np.shape[0])))
+@pytest.mark.parametrize('sigma', (constr, covariance, constr_tf))
 def test_unbinned_nll(weights, sigma):
     gaussian1, mu1, sigma1 = create_gauss1()
     gaussian2, mu2, sigma2 = create_gauss2()
--- a/tests/test_loss.py
+++ b/tests/test_loss.py
@@ -53,7 +53,7 @@ mu_constr = [1.6, 0.2]  # mu, sigma
 sigma_constr = [3.8, 0.2]
 constr = lambda: [mu_constr[1], sigma_constr[1]]
 constr_tf = lambda: ztf.convert_to_tensor(constr())
-covariance = lambda: np.array([[mu_constr[1]**0.5, -0.05], [-0.05, sigma_constr[1]**0.5]])
+covariance = lambda: np.array([[mu_constr[1] ** 0.5, -0.05], [-0.05, sigma_constr[1] ** 0.5]])
 covariance_tf = lambda: ztf.convert_to_tensor(covariance())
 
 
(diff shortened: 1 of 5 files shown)


zfit/constraint.py
```diff
@@ -5,7 +5,8 @@
 
 __all__ = ["nll_gaussian"]
 
-#def nll_gaussian(params, mu, sigma):
+
+# def nll_gaussian(params, mu, sigma):
```
apuignav: I would say just remove the old one? We see the changes anyway in git :-)

marinang: True.

marinang:
A test with a covariance matrix is missing, I will add a 2D example.

marinang:
🤔 I am not sure why the build fails, is it because of `iminuit` ?

mayou36:
[yes](https://github.com/scikit-hep/iminuit/issues/337)
well see once iminuit is fixed

zfit/constraint.py
```diff
@@ -1,19 +1,53 @@
...
+        constraint
+    Returns:
+        graph: the nll of the constraint
+    Raises:
+        ValueError: if params, mu and sigma don't have the same size
+    """
+
+    params = convert_to_container(params, tuple)
+    mu = convert_to_container(mu, tuple)
+
```
mayou36: The function should in general accept tensors, python and numpy objects being the a special case of a constant tensor.
This makes things of course a little bit more complicated :)
params and mu have to be scalars (probably allow mu as tensor later on), we leave it like that. But make sure that sigma can be a tensor instead of a numpy array. So let's do all the conversion to Tensor as soon as possible and use all tf operations (tf.diag) to do the rest.

This should also remove to special case for `isnumber`, since it's just the special case of a Tensor with n axes.

You agree? Let me know if you get stuck with TF somewhere

tests/test_loss.py
```diff
@@ -107,8 +108,10 @@ def test_unbinned_simultaneous_nll():
 
 
 @pytest.mark.flaky(3)
-@pytest.mark.parametrize('weights', [None, np.random.normal(loc=1., scale=0.2, size=test_values_np.shape[0])])
-def test_unbinned_nll(weights):
+@pytest.mark.parametrize('weights, sigma', [(None, [mu_constr[1], sigma_constr[1]]), (None, covariance),
+                         (np.random.normal(loc=1., scale=0.2, size=test_values_np.shape[0]),
+                         [mu_constr[1], sigma_constr[1]])])
+def test_unbinned_nll(weights, sigma):
```
mayou36: (sorry for my slow reaction recently, been quite busy)
Can you do two changes? I think then we're good

- separate the parametrize? They can simply be stacked. So that there is a weight parameterization and a gaussian constraint parametrisation.
- can you make the constraints callables? (and in the code just call them?) This allows to also have tensors in the test, since they cannot be build yet outside the function (e.g. in the parametrize) and add a tensor as test? (can be simply: lambda: tf.constant(mu_constr[0])

(not for _now_, this works of course, but for future that we don't break things)

marinang: I don't understand why making the constraints callable?

mayou36: @marinang according to my email, you posted something, asking why the callable is needed. I don't see it on github. Weird...
Anyway, the reason is that the parametrization of the test happens at collection time, so an arbitrary time _before_ the test is executed. Since, due to the setup_function, the graph is reset before each test (to have clean tests). Since a tensor is added to the graph when the method is called (e.g. `tf.constant` this would add the constant to the graph at collection time. This gets reset and at the test runtime, this would point to an object from a different (because reset) graph. With the callable, we make sure that the actual op gets added to the right graph.

On the parametrization: you can remove most of them, it is the product of combinations, a matrix of the possible combinations.

so:
```
parametrized("weights", [None, something])
parametrized("other", [a, b, c])
```
gives 2 x 3 combinations of tests that are run.

The CI is failing, I think due to an indentation error. Typo somewhere?

marinang: Ok great I just discovered very recently the parametrize option of pytest, nice feature. For the failure of “pr” Travis build, there is a lot of red everywhere and it’s difficult for me to catch where the error comes from (also not easy to read the log on a phone).

mayou36: I agree, I don't see it either. Anyway, I fix the pytest parametrization and the bug

PR #155, opened 2019-06-20
Add covariance method the FitResult.

marinang:
## Proposed Changes

Add covariance method in `FitResult`. Takes `params` as argument, and `as_dict` to return a dictionary.

## Tests added

None yet, Is there a place where the `FitResult` ( I didn't find it.) ? If not I will just create a test file.

### DIFF
--- a/tests/test_fitresult.py
+++ b/tests/test_fitresult.py
@@ -5,6 +5,7 @@ from zfit.core.testing import setup_function, teardown_function, tester
 import tensorflow as tf
 import zfit
 from zfit import ztf
+import numpy as np
 
 true_a = 1.
 true_b = 4.
@@ -55,7 +56,10 @@ def create_fitresult(minimizer_class_and_kwargs):
     cur_val = zfit.run(loss.value())
     aval, bval, cval = zfit.run([v for v in (a_param, b_param, c_param)])
 
-    return {'result': result, 'true_min': true_minimum, 'cur_val': cur_val, 'a': aval, 'b': bval, 'c': cval}
+    ret = {'result': result, 'true_min': true_minimum, 'cur_val': cur_val, 'a': aval, 'b': bval, 'c': cval,
+           'a_param': a_param, 'b_param': b_param, 'c_param': c_param}
+
+    return ret
 
 
 minimizers = [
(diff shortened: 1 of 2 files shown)


tests/test_fitresult.py
```diff
@@ -0,0 +1,78 @@
...
+@pytest.mark.parametrize("minimizer_class_and_kwargs", minimizers)
+def test_fmin(minimizer_class_and_kwargs):
+    results = create_fitresult(minimizer_class_and_kwargs=minimizer_class_and_kwargs)
+    result = results['result']
+    assert pytest.approx(results['cur_val']) == result.fmin
+
+
+@pytest.mark.parametrize("minimizer_class_and_kwargs", minimizers)
+def test_covariance(minimizer_class_and_kwargs):
+    pass
```
mayou36: I've added an explicit fitresult test file, feel free to put any test here. As an exapmle, see the `test_fmin` and the return from line 58

zfit/minimizers/fitresult.py
```diff
@@ -239,6 +242,50 @@ def _error(self, params, method, sigma):
...
+                params (list(:py:class:`~zfit.Parameter`)): The parameters to calculate
+                    the covariance matrix. If `params` is `None`, use all *floating* parameters.
+                as_dict (bool): Default `False`. If `True` then returns a dictionnary.
+
+            Returns:
+                2D `numpy.array` of shape (N, N);
+                `dict`(param1, param2) -> covariance if `as_dict == True`.
+        """
+        params = self._input_check_params(params)
+        covariance_dict = self.minimizer._minuit_minimizer.covariance
```
mayou36: We need to be independent of minuit for sure. So for the moment, we can use the minuit covariance, but a standalone function should be there at some point (using _any_ library, we can use pure np for the calulations as well)

Can you put it into a try-except block catching the `AttributeError` and saying that only minuit covariance is currently there?

zfit/minimizers/fitresult.py
```diff
@@ -239,6 +242,50 @@ def _error(self, params, method, sigma):
...
+def dict_to_matrix(params, matrix_dict):
+
+    nparams = len(params)
+    matrix = np.empty((nparams, nparams))
+
+    for i in range(nparams):
+        pi = params[i]
+        for j in range(i, nparams):
+            pj = params[j]
+            key = (pi, pj)
```
mayou36: I've refactored things slightly to always use the `Parameter` objects and not the string names.

marinang: Good, it makes sense to use the `Parameter`.

zfit/minimizers/fitresult.py
marinang: Ok I agree but `MinuitMinimizer`doesn't have a covariance method. What's your idea of covariance implementation with other minimiser?

mayou36: Something similar to minuit for example. If I understand that correctly, it's an inversion of the error matrix, removal of the line and column that contains the parameter and a re-inversion.

Getting the error matrix, by hessian or by minos-like profiling, is another story, but also doable

marinang: Yeah it's something like this, like the inverse of the hessian matrix of the log-likelihood.

marinang:
This should be good. The build failure comes from `test_complex_param`.

PR #156, opened 2019-06-22
Add new constraint classes

marinang:
Creation of constraints object

## Proposed Changes

  - Creation of `BaseConstraint` depending only on `BaseNumeric`
  - Creation of `SimpleConstraint` that can take tensor
  - Creation of `DistributionConstraint` that can tfp pdfs, samples of constrained parameters possible
  - Creation of `GaussianConstraint` depending on `DistributionConstraint` 

## Tests added

Modification of tests in `test_loss.py` and `test_constraint`.

### DIFF
--- a/tests/test_constraint.py
+++ b/tests/test_constraint.py
@@ -6,7 +6,7 @@ import zfit
 from zfit import ztf
 from zfit.core.testing import setup_function, teardown_function, tester
 from zfit.util.exception import ShapeIncompatibleError
-from zfit.constraint import SimpleConstraint, GaussianConstraint
+from zfit.core.constraint import BaseConstraint, SimpleConstraint, GaussianConstraint
 from zfit.util.container import convert_to_container
 
 
@@ -23,7 +23,13 @@ def nll_gaussian(params, mu, sigma):
     return constraint
 
 
-def test_shape_errors():
+def test_base_constraint():
+    constr = BaseConstraint()
+    with pytest.raises(NotImplementedError):
+        constr.value()
+
+
+def test_gaussian_constraint_shape_errors():
     param1 = zfit.Parameter("Param1", 5)
     param2 = zfit.Parameter("Param2", 6)
 
@@ -37,7 +43,7 @@ def test_shape_errors():
         GaussianConstraint(param1, mu=[4, 2], sigma=[2, 3])
 
 
-def test_gaussian_constraint_value():
+def test_gaussian_constraint():
     param1 = zfit.Parameter("Param1", 5)
     param2 = zfit.Parameter("Param2", 6)
     params = [param1, param2]
@@ -48,6 +54,8 @@ def test_gaussian_constraint_value():
     constr = GaussianConstraint(params=params, mu=mu, sigma=sigma)
     constr_np = zfit.run(constr.value())
     assert constr_np == pytest.approx(3.989638)
+
+    assert constr.get_dependents() == set(params)
 
 
 def test_GaussianConstraint_sampling():
@@ -79,3 +87,5 @@ def test_simple_constraint():
 
     constr_np = zfit.run(constr.value())
     assert constr_np == pytest.approx(2.02)
+
+    assert constr.get_dependents() == set(params)
--- a/tests/test_constraint.py
+++ b/tests/test_constraint.py
@@ -8,6 +8,8 @@ from zfit.core.testing import setup_function, teardown_function, tester
 from zfit.util.exception import ShapeIncompatibleError
 from zfit.core.constraint import BaseConstraint, SimpleConstraint, GaussianConstraint
 from zfit.util.container import convert_to_container
+import tensorflow_probability as tfp
+poisson = tfp.distributions.poisson
 
 
 def nll_gaussian(params, mu, sigma):
--- a/tests/test_constraint.py
+++ b/tests/test_constraint.py
@@ -60,8 +60,8 @@ def test_GaussianConstraint_sampling():
 
     sample = zfit.run(constr.sample(15000))
 
-    assert np.mean(sample[param1]) == pytest.approx(5, rel=0.01)
-    assert np.std(sample[param1]) == pytest.approx(1, rel=0.01)
+    assert np.mean(sample[param1]) == pytest.approx(mu[0], rel=0.01)
+    assert np.std(sample[param1]) == pytest.approx(sigma[0], rel=0.01)
 
 
 def test_simple_constraint():
--- a/tests/test_constraint.py
+++ b/tests/test_constraint.py
@@ -8,8 +8,6 @@ from zfit.core.testing import setup_function, teardown_function, tester
 from zfit.util.exception import ShapeIncompatibleError
 from zfit.core.constraint import BaseConstraint, SimpleConstraint, GaussianConstraint
 from zfit.util.container import convert_to_container
-import tensorflow_probability as tfp
-poisson = tfp.distributions.poisson
 
 
 def nll_gaussian(params, mu, sigma):
--- a/tests/test_constraint.py
+++ b/tests/test_constraint.py
@@ -6,6 +6,21 @@ import zfit
 from zfit import ztf
 from zfit.core.testing import setup_function, teardown_function, tester
 from zfit.util.exception import ShapeIncompatibleError
+from zfit.constraint import SimpleConstraint, GaussianConstraint
+from zfit.util.container import convert_to_container
+
+
+def nll_gaussian(params, mu, sigma):
+    params = convert_to_container(params, container=tuple)
+    mu = convert_to_container(mu, container=tuple)
+    sigma = convert_to_container(sigma, container=tuple)
+    constraint = ztf.constant(0.)
+    if not len(params) == len(mu) == len(sigma):
+        raise ValueError("params, mu and sigma have to have the same length.")
+    for param, mean, sig in zip(params, mu, sigma):
+        constraint += ztf.reduce_sum(ztf.square(param - mean) / (2. * ztf.square(sig)))
+
+    return constraint
 
 
 def test_shape_errors():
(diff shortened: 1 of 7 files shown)


zfit/core/constraint.py
```diff
@@ -0,0 +1,178 @@
...
+
+        Args:
+            distribution (tensorflow_probability.distributions.Distribution): The probability density function
+                used to constraint the parameters
+
+        """
+
+        super().__init__(params=params, name=name, dtype=dtype, **kwargs)
+
+        self._distribution = distribution
```
mayou36: This is a very tricky one: https://github.com/tensorflow/probability/issues/285
In short, for on-the-fly computations of the normalization (a feature actually  working but not yet extensively tested), we need the `GradientTape`. And this only works, if the distribution is _created while invoking the pdf_.

So what we need is something like in https://github.com/zfit/zfit/blob/ea2f6446a090a55e4572bd3a70ebdc068772a0a1/zfit/models/dist_tfp.py#L56 _for the value_ (for the `sample` it is not actually needed, but let's put it as well).

What is needed in the `__init__` is an argument for the instatiation_params _and_ instantiation args. While I agree that _most probably_ there is no usecase for a changeable parametrization of the Constraint, it still is a cleaner separation.

zfit/core/loss.py
```diff
@@ -59,6 +60,19 @@ def _nll_constraints_tf(constraints):
     return constraints_neg_log_prob
 
 
+def _constraint_check(constraints):
+    _constraints = []
+    for c in constraints:
```
mayou36: c -> constr or constraint  (rather no single letter names)

zfit/core/loss.py
```diff
@@ -59,6 +60,19 @@ def _nll_constraints_tf(constraints):
     return constraints_neg_log_prob
 
 
+def _constraint_check(constraints):
+    _constraints = []
```
mayou36: `_constraints` -> `checked_constraints`?

zfit/core/loss.py
```diff
@@ -59,6 +60,19 @@ def _nll_constraints_tf(constraints):
     return constraints_neg_log_prob
 
 
+def _constraint_check(constraints):
```
mayou36: add `convert` into the name? Since it's not only a check

tests/test_constraint.py
```diff
@@ -6,30 +6,86 @@
 from zfit import ztf
 from zfit.core.testing import setup_function, teardown_function, tester
 from zfit.util.exception import ShapeIncompatibleError
+from zfit.core.constraint import BaseConstraint, SimpleConstraint, GaussianConstraint
+from zfit.util.container import convert_to_container
 
 
-def test_shape_errors():
+def nll_gaussian(params, mu, sigma):
```
mayou36: Probably -> `true_nll_gaussian`?

marinang:
Good. I will work on the few points you stated and suggestions.

> The whole Constraint look now very similar to the normal PDF class. There are good reasons to keep them separate (e.g. you don't do products with Constraints, no integrals etc.), but the rational of what exactly distinguishes them should be clear. Just to keep in mind.

Yes I agree, funny enough in `RooFit` this is how people are doing, mathematically it's not wrong but can be misleading.

> For the SimpleConstraint, we can even use (at some point, no need for now) the accept_reject_sampling function :)

I have to check what this function does. 

> the name: sample is already used in the model, but has a different API here, so just "not optimal", most of all if it return a numpy array instead of a tensor (as PDF.sample does0. Any alternative name? But not crucial...

I understand your concern, I don't have alternative name in mind for now.

Concerning "How to use it" I also prefer the first way.

mayou36:
> Yes I agree, funny enough in `RooFit` this is how people are doing, mathematically it's not wrong but conceptually it is.

yep, let's see if we can successfully keep the two separated in zfit.

> > For the SimpleConstraint, we can even use (at some point, no need for now) the accept_reject_sampling function :)
> 
> I have to check what this function does.

Probably to complicated with the shapes and dependents, we'll see. But anyway, not important now. It simply does accept reject sampling. But maybe an mcmc sampler or similar. We'll see!

> I understand your concern, I don't have alternative names in mind for now.

Fine, let's keep it for now then

> Concerning "How to use it" I also prefer the first way.

Also that it returns a numpy array?

One thing I forgot: the limits of a parameter. If the sampled value is not inside the limits, we _could_ resample _or_ say that it is up to the user to specify the Constraint right. I think the first is preferable, though actually not correct (since the PDF should then be normalized over another range, in principle), right?

marinang:
Yep returning numpy arrays. But are you ok with the returned  format Dict(Params, array) ?

Regarding the limits of `Parameter` isn't it what `accept_reject_sample` is supposed to do?

zfit/core/constraint.py
```diff
@@ -85,14 +78,14 @@ def _get_dependents(self):
...
 
 
-class ProbConstraint(BaseConstraint):
+class DistributionConstraint(BaseConstraint):
 
-    def __init__(self, params: Dict[str, ZfitParameter], distribution: Distribution,
-                 name: str = "ProbConstraint", dtype=ztypes.float,
+    def __init__(self, params: Dict[str, ZfitParameter], distribution: tfd.Distribution,
+                 dist_params, dist_kwargs=None, name: str = "DistributionConstraint", dtype=ztypes.float,
                  **kwargs):
```
mayou36: Default missing: `if dist_kwargs is None` use `{}` (otherwise fails in the distribution property)

marinang: You would rather have `dist_kwargs={}` in the function arguments or inside the function
```
if dist_kwargs is None:
    dist_kwargs = {}
```

zfit/core/constraint.py
```diff
@@ -128,10 +129,10 @@ def sample(self, n: ztyping.nSamplingTypeIn = 1) -> Dict:
         return {p: sample[:, i] for i, p in enumerate(self.get_params())}
 
     def _sample(self, n):
-        return self.distribution.sample(n)
+        return zfit.run(self.distribution.sample(n))
```
mayou36: We need to cache that in the future at least. Currently, the caching is not optimal and should be easier. Can you just mark it with a `# TODO cache: add proper caching`?

The drawback is not hard, just that _if used inside a loop_, it will become slow. But that's anyway not how it should be used.

zfit/core/constraint.py
mayou36: Inside is fine, as there is always the issue of the mutable default arg.
So let's keep None and convert inside, agree?

mayou36:
> Yep returning numpy arrays. But are you ok with the returned format Dict(Params, array) ?

yes, absolutely, very clean!

> Regarding the limits of `Parameter` isn't it what a `accept_reject_sample` function is supposed to do?

Not quite, this is more for a pdf and sampling in the data, not the parameters. And the limits (to be sampled from) and norm_range are there handled separately anyway.

what we need is an analytic sampling from cdf and icdf (that are both available with `Distribution`), it's actually anyway needed for the pdfs and already there more or less. I can take care of that, so let's ignore for the moment the limits, okay?

marinang:
So ok maybe one last thing to do is to add a word about sampling of the constraint in the `model.rst`.

mayou36:
Good! I'll check in the next two days whether we can put the analytic sampling there easily, if so I'll add it, otherwise we just merge.

Still a conceptual thing though: given limits, should we change the normalization to reflect the limits? It's in fact very easy to do with `Distribution` (divide the prob by cdf(upper) - cdf(lower)). I just think that _if_ we limit the sampling for the limits, we _may_ should also alter the pdfs normalization. It seems inconsistent otherwise. What do you think?

zfit/core/constraint.py
```diff
@@ -0,0 +1,183 @@
...
+        return self._value()
+
+    @abc.abstractmethod
+    def _value(self):
+        raise NotImplementedError
+
+    def _get_dependents(self) -> ztyping.DependentsType:
+        return self._extract_dependents(self.get_params())
+
+    def sample(self, n):
```
mayou36: no default for n, because n=1 is rather a special case than the vast majority of the cases (right?). So rather require it

zfit/core/constraint.py
```diff
@@ -0,0 +1,183 @@
...
+        kwargs = self.dist_kwargs
+        if callable(kwargs):
+            kwargs = kwargs()
+        return self._distribution(**params, **kwargs, name=self.name + "_tfp")
+
+    def _value(self):
+        value = -self.distribution.log_prob(self._tparams)
+        return value
+
+    def _sample(self, n):
```
mayou36: moved the public `sample` to the base class

zfit/core/constraint.py
marinang: Ok 👍🏻

zfit/core/constraint.py
```diff
@@ -0,0 +1,191 @@
...
+        raise NotImplementedError
+
+
+class SimpleConstraint(BaseConstraint):
+
+    def __init__(self, func: Callable, params: Optional[ztyping.ParametersType] = None, sampler: Callable = None):
+        """Constraint from a (function returning a) Tensor.
+
+        The parameters are named "param_{i}" with i starting from 0 and corresponding to the index of params.
```
marinang: Why the parameters need to be renamed?

mayou36: Sorry for the slow reaction!

Let's put it the other way: parameter.name as key is circular. Assuming that _whenever you have the name you also have the object -> no access with name strings_:
to get a parameter, you need the parameter itself then
```
param1 = constr.params[param1.name]
```
which is kinda useless. Furthermore, if you _don't_ know, you cannot access the parameter. But the param names should be the names of the _parametrization_, not of the actual value used. You get that by accessing the parametrization.

And `"param_{i}"` is just a generic name you can access easily programatically. But it's _actually_ not really needed or important.

Do you agree?

marinang: Ok yes I agree :)

PR #158, opened 2019-06-27
Changes in doc for the new constraint classes

marinang:
I fixed typos and I changed parts that didn't look nice in the doc for the constraints.

See https://zfit.github.io/zfit/model.html, and https://zfit.github.io/zfit/loss.html.

### DIFF
--- a/docs/loss.rst
+++ b/docs/loss.rst
@@ -84,5 +84,5 @@ The same result can be achieved by passing a list of PDFs on instantiation, alon
 
 .. code-block:: pycon
 
-   >>> # Adding a list of models, data and observable ranges
+   >>> # Adding a list of models and datasets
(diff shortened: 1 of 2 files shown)


apuignav:
Nice catch! I was wondering if you could include a small example on how to actually write a custom constraint. What do you think?

marinang:
@apuignav like from function that returns a Tensor? Do you have a good constraint example?

apuignav:
I was thinking even explaining a bit the `GaussianConstraint`, but it's actually quite complex. What about a simple parameter inequality constraint? Something like: if the parameter value is above a certain value, make the logL explode. This would be extremely useful for setting limits on composed parameters, for example.

marinang:
Good I was thinking about something similar. I'll write the example.

mayou36:
Looks nice! As soon as you wanna merge, just request the review.

PR #167, opened 2019-09-12
Fix gauss constr

### LINKED ISSUE #162: Bug with GaussianConstraint
I want to add 3 constraints to a likelihood, see dummy example:

```python

const1 = zfit.Parameter("const1",  1.0)
const2 = zfit.Parameter("const2",  2.0)
const3 = zfit.Parameter("const3",  3.0)

constraint = {"params": [const1, const2, const3],
                        "mu": [1.0, 2.0, 3.0],
                        "sigma": [0.05, 0.05, 0.05]}

nll.add_constraints(zfit.constraint.GaussianConstraint(**constraint))

```
So this works but when I minimise the likelihood the parameters are not even close the the central value of the parameter and even if I put limits to the parameters this doesn't change anything (or the parameters are a the limits after the minimisation).

On the other hand If I do 

```python

const1 = zfit.Parameter("const1",  1.0)
const2 = zfit.Parameter("const2",  2.0)
const3 = zfit.Parameter("const3",  3.0)

constraint = {"params": [const1, const2],
                        "mu": [1.0, 2.0],
                        "sigma": [0.05, 0.05]}

nll.add_constraints(zfit.constraint.GaussianConstraint(**constraint))
nll.add_constraints(zfit.constraint.GaussianConstraint(params=const3, mu=const3, sigma=0.05))
```
It works perfectly ... the parameters are very close the central values of the constraints.
I have not clue yet why this happens.

## Context (Environment)

* zfit version: 0.3.3
* Python version: 3.6.7
* Are you using conda, pipenv, etc? : Conda
* Operating System: MacOS
* Tensorflow version: 1.13.1

mayou36:
Fixes #162 

## Proposed Changes

  -

## Tests added

  - test for three parameters and Gaussian constraint
  
## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated

### DIFF
--- a/tests/test_constraint.py
+++ b/tests/test_constraint.py
@@ -83,7 +83,7 @@ def test_gaussian_constraint():
     assert constr2_np == pytest.approx(true_val2)
 
 
-def test_gaussian_constraint_bug():
+def test_gaussian_constraint_orderbug():  # as raised in #162
     param_vals = [1500, 1.0, 1.0, 1.0, 0.5]
     params = [zfit.Parameter(f"param{i}", val) for i, val in enumerate(param_vals)]
 
@@ -99,15 +99,15 @@ def test_gaussian_constraint_bug():
     assert true_val < 10000
 
 
-def test_gaussian_constraint_bug2():
-    luminosity = zfit.Parameter("lumi", 1500)
-    efficiency = zfit.Parameter("eff", 0.5)
+def test_gaussian_constraint_orderbug2():  # as raised in #162, failed before fixing
+    param1 = zfit.Parameter("param1", 1500)
+    param5 = zfit.Parameter("param2", 0.5)
 
-    syst1 = zfit.Parameter("syst1", 1.0)
-    syst2 = zfit.Parameter("syst2", 1.0)
-    syst3 = zfit.Parameter("syst3", 1.0)
+    param2 = zfit.Parameter("param3", 1.0)
+    param3 = zfit.Parameter("param4", 1.0)
+    param4 = zfit.Parameter("param5", 1.0)
 
-    constraint = {"params": [luminosity, syst1, syst2, syst3, efficiency],
+    constraint = {"params": [param1, param2, param3, param4, param5],
                   "mu": [1500, 1.0, 1.0, 1.0, 0.5],
                   "sigma": [0.05 * 1500, 0.001, 0.01, 0.1, 0.05 * 0.5]}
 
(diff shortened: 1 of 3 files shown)


tests/test_constraint.py
```diff
@@ -58,6 +59,76 @@ def test_gaussian_constraint():
...
+
+    constraint = {"params": [luminosity, syst1, syst2, syst3, efficiency],
+                  "mu": [1500, 1.0, 1.0, 1.0, 0.5],
+                  "sigma": [0.05 * 1500, 0.001, 0.01, 0.1, 0.05 * 0.5]}
+
+    constr1 = zfit.constraint.GaussianConstraint(**constraint)
+    # param_vals = [1500, 1.0, 1.0, 1.0, 0.5]
+    constraint['params'] = zfit.run(constraint['params'])
+
+    true_val = true_gauss_constr_value(**constraint)
```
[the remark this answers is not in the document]
mayou36: no, bug2 errors...

mayou36:
> It also seems that the order of the params, mu and sigma list is not preserved as shown in the example I sent you. Maybe a test checking that this order is kept can be added.

This would explain a lot, but I do not see where it gets messed up or how...

zfit/core/constraint.py
```diff
@@ -184,10 +183,11 @@ def __init__(self, params: ztyping.ParamTypeInput, mu: ztyping.NumericalScalarTy
 
         self._covariance = covariance
         self._mu = mu
+        self._params_array = params
 
         distribution = tfd.MultivariateNormalFullCovariance
         dist_params = dict(loc=mu, covariance_matrix=covariance)
-        dist_kwargs = dict(validate_args=True)
+        dist_kwargs = dict(validate_args=False)
```
marinang: Why remove this check except for possible improvement in runtime performance?

mayou36: Good catch, thx! Forgot to change that back (done now)

mayou36:
So I found it, and added a fix. I am not fully happy, maybe we need to restructure the Constraints anyway (e.g. also name? Make sure it's clear that it's an NLL constraint? But other topic)

mayou36:
Ok, fixed, @marinang can you approve if ok?

zfit/core/constraint.py
```diff
@@ -124,7 +124,6 @@ def __init__(self, params: Dict[str, ZfitParameter], distribution: tfd.Distribut
         self._distribution = distribution
         self.dist_params = dist_params
         self.dist_kwargs = dist_kwargs if dist_kwargs is not None else {}
-        self._tparams = tf.stack(self.get_params())
```
marinang: So this was the source of the problem, so `get_params` doesn't conserve the order ?

mayou36: Yes, not insertion order at least, not guaranteed...

PR #184, opened 2019-12-08
moving to TF2 eager

mayou36:
## Proposed Changes

 - Make zfit fully using TensorFlow 2.0 with eager mode, functions etc
 - breaking changes in things containing dependencies to parameters as we cannot rely on the graph anymore
 - in general: way more written in "numpy" style

  
## Checklist
 - [x] fix constraints to make them depend on sigma
 - [x] add upgrade errors
 - [x] create upgrade guide
 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated


docs/getting_started.rst
```diff
@@ -145,6 +145,28 @@ One example is the Gauss PDF that has been shown above. The object ``gauss`` con
...
+    >>> probs = gauss.pdf(consts, norm_range=(-np.infty, np.infty))
+
+    >>> # And now execute the tensorflow graph
+    >>> result = zfit.run(probs)
+    >>> print("x values: {}\nresult:   {}".format(consts, result))
+    x values: [-1, 0, 1]
+    result:   [0.24262615 0.39670691 0.24130008]
+
+Integrating a given PDF for a given normalisation range also returns a graph, so it needs to be run using
+    >>> from zfit import z
```
apuignav: Why this import? You don't use `z` right?

mayou36: I changed the name from `ztf` to `z`, basically the wrapped TensorFlow. `ztf` seems to be too close to `zfit`, so it causes confusion.

But you're right, we don't use it here. And yep, need to cleanup the docs anyway, this was autoconversion I guess.

Some things look slightly different now in TF2, I'll write an upgrade guide as well

PR #197, opened 2020-01-29
fix error message in core/loss.py

niklasnolte:
untested

unrelated (and unimportant), but i'd not use a double negation in anything (in this case `IntentionNotUnambiguousError` -> `IntentionAmbiguousError`)

### DIFF
--- a/zfit/core/loss.py
+++ b/zfit/core/loss.py
@@ -135,10 +135,10 @@ class BaseLoss(BaseDependentsMixin, ZfitLoss, Cachable, BaseObject):
             fit_range = []
             for p, d in zip(pdf, data):
                 if not p.norm_range == d.data_range:
-                    raise IntentionNotUnambiguousError("No `fit_range` is specified and `pdf` {} as "
-                                                       "well as `data` {} have different ranges they"
-                                                       "are defined in. Either make them (all) consistent"
-                                                       "or specify the `fit_range`")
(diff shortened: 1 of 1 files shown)


mayou36:
Nice, thanks! The idea of this double negotiation is to but emphasize that it has to be Unambiguous. But yeah, we could also reformulate that, I'll keep it in mind

mayou36:
Looks good so far. Can I merge?

and since you already start contributing (I already wrote in a different place, maybe you also missed it, just to be sure): are you sure you are not interested to become part of the project? There are many different levels of participation, also smaller, simpler parts, but also larger, or just for discussions. If you are at CERN anytime soon (you're stationed there?), we can have a chat over a coffee and I can show you the library and project in more detail. Just in case you are interested and without any commitment, it is also useful for us to know the user-base and what projects zfit is used for.

In case, you can also contact me directly at Jonas.Eschle@cern.ch

niklasnolte:
i don't think i will be fitting an awful lot in the near future, but whenever i do, i'll happily report and help to fix all problems or bugs that may arise. fyi, i am currently using zfit to build a lab exercise where students explore classification strategies in physics analyses. good work so far, keep it up.

mayou36:
That's good to know, thanks. And feel free to point interested students to us, I have from time to time students that find it quite interesting to work on a large software project.

Thanks again for the contribution and fixing of things!

PR #200, opened 2020-02-19
Update index.rst

marinang:
Replace lauztat by hepstats.


mayou36:
I've changed to develop, we can anyway release very soon 0.4.2 to have the new constraints in there

PR #202, opened 2020-02-21
Cb test fail

mayou36:
Fixes #

## Proposed Changes

  -
  -
  -

## Tests added

  -
  -
  -
  
## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated

### DIFF
--- a/tests/test_pdf_cb.py
+++ b/tests/test_pdf_cb.py
@@ -8,8 +8,8 @@ import zfit
 from zfit.core.testing import setup_function, teardown_function, tester
 from zfit.models.physics import CrystalBall, DoubleCB
 
-mu = 0.0
-sigma = 0.5
+mu = -0.2
+sigma = 0.7
 
 alphal = 1.0
 nl = 2.0
--- a/tests/test_pdf_cb.py
+++ b/tests/test_pdf_cb.py
@@ -8,16 +8,16 @@ import zfit
 from zfit.core.testing import setup_function, teardown_function, tester
 from zfit.models.physics import CrystalBall, DoubleCB
 
-mu = -0.2
-sigma = 0.7
+mu = -0.3
+sigma = 1.1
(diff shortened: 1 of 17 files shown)


tests/test_pdf_cb.py
```diff
@@ -91,7 +91,7 @@ def test_cb_dcb():
     assert not any(np.isnan(probsr))
 
     probsl_scipy = crystalball.pdf(x, beta=alphal, m=nl, loc=mu, scale=sigma)
-    probsr_scipy = crystalball.pdf(-x, beta=alphar, m=nr, loc=mu, scale=sigma)
+    probsr_scipy = crystalball.pdf(-x + 2 * mu, beta=alphar, m=nr, loc=mu, scale=sigma)
```
marinang: Is this dark magic :D ?

mayou36: Haha kinda: we want to mirror it at the mu. x is not at mu, we mirror it (now at -mu), shift it back.

Simple math thing, mirroring at point mu

mayou36:
Sorry for already merged, did not expect you to have time now ;) Let me know if you have another comment, there have been a few changes (see updated changelog)

tests/test_pdf_cb.py
marinang: Oh yeah cause mu is not equal to zero anymore.

mayou36: yep, just breaking some symmetries to see things still work

marinang:
@mayou36 It is fine no worries, I was just curious about this PR since I also was somehow affected by the CB a while ago.

mayou36:
Ah yep, solutions: use the new option "use_minuit_grad" in Minuit to use the numerical gradient, then things work out better

PR #203, opened 2020-02-26
WIP: Space refactoring

mayou36:
Fixes #

## Proposed Changes

  -
  -
  -

## Tests added

  -
  -
  -
  
## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated


zfit/core/data.py
```diff
@@ -362,23 +360,24 @@ def numpy(self):
...
+            #         upper = tf.cast(tf.transpose(upper), dtype=self.dtype)
+            #         lower = tf.cast(tf.transpose(lower), dtype=self.dtype)
+            #
+            #     below_upper = tf.reduce_all(input_tensor=tf.less_equal(value, upper), axis=1)  # if all obs inside
+            #     above_lower = tf.reduce_all(input_tensor=tf.greater_equal(value, lower), axis=1)
+            #     inside_limits.append(tf.logical_and(above_lower, below_upper))
+            # inside_any_limit = tf.reduce_any(input_tensor=inside_limits, axis=0)  # has to be inside one limit
+            #
+            # value = tf.boolean_mask(tensor=value, mask=inside_any_limit)
             # value = tf.transpose(value)
```
SebastianJL:
```suggestion
```

SebastianJL:
I wouldn't call the new file `space_new.py`. As far as I can see `limits.Space` was only exposed to the user in `zfit.__init__`. Just rename `limits.py` to 'space.py'.

mayou36:
yep, sure. And it will be named space as well, most probably. It's more a relict of the refactoring. And maybe we need to keep the old space around or something for another release.

tests/test_limit.py
```diff
@@ -0,0 +1,139 @@
...
+    if graph:
+        test = z.function(test)
+    inside, inside2, outside, equal, inside21, inside22, outside2 = test()
+    assert not equal
+    assert inside
+    assert inside2
+    assert not outside
+    assert inside21
+    assert inside22
+    assert not outside2
```
SebastianJL: I'll make this comment only for this test case but it applies to all test cases in this file.
Why do you split this into a function `test()` and a block of assertions (line 133-139)?

If one of the assertions fails, I have to search first search for the related code to find it.

Just write the assertion on the line where you would assign the variable. Most of your expression are self-explanatory and don't need to be named by a variable. Example:
Do
```python
assert limit.inside(2)
```
instead of 
```python
inside = limit.inside(2)
# ... twenty lines of code
assert inside
```
This way you can replace the function `test()` by it's body, i.e. delete about 15 lines of code and make it more concise. Do not try to generalize stuff in advance! One of the best guidelines I learnt in my internship is the rule of three. Only refactor stuff into a function (or other general procedure) once you encounter the same code three 
times. Otherwise your make premature assumptions of what needs to be generalized, which ultimately leads to more refactoring and more code. 
https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

zfit/__init__.py
```diff
@@ -68,15 +68,18 @@ def _maybe_disable_warnings():
 
 from . import constraint, pdf, minimize, loss, core, data, func, param
 from .core.parameter import Parameter, ComposedParameter, ComplexParameter, convert_to_parameter
-from .core.limits import Space, convert_to_space, supports
+from .core.space import Space, convert_to_space, supports
+# from .core.spaces import Space, convert_to_space, supports
```
SebastianJL:
```suggestion
```

zfit/__init__.py
```diff
@@ -68,15 +68,18 @@ def _maybe_disable_warnings():
...
 
 from .settings import run
 
 
 def _maybe_disable_jit():
     import os
-    z.zextension.FunctionWrapperRegistry.do_jit = bool(int(os.environ.get("ZFIT_DO_JIT", True)))
+    z.zextension.FunctionWrapperRegistry.do_jit = bool(int(os.environ.get("ZFIT_DO_JIT", True))) and bool(
+        int(os.environ.get("ZFIT_EXPERIMENTAL_DO_JIT", True)))
+    z.zextension.FunctionWrapperRegistry2.do_jit = z.zextension.FunctionWrapperRegistry.do_jit
```
SebastianJL: See comments on `zfit/z/zextension.py`

zfit/core/coordinates.py
```diff
@@ -0,0 +1,309 @@
...
+        if self.obs is not None and other.obs is not None:
+            obs_equal = frozenset(self.obs) == frozenset(other.obs)
+
+        if self.axes is not None and other.axes is not None:
+            axes_equal = frozenset(self.axes) == frozenset(other.axes)
+        equal = obs_equal or axes_equal
+        return equal
+
+    def __hash__(self):
+        return 42  # always check with equal...  maybe change in future, use dict that checks for different things.
```
SebastianJL: How is this useful? Why implement `__hash__()` if the object is not hashable?

zfit/core/dimension.py
```diff
@@ -181,108 +126,115 @@ def common_obs(spaces: ztyping.SpaceOrSpacesTypeInput) -> List[str]:
...
+#     new_upper = check_extract_limits(all_upper)
+#     assert not (new_lower is None) ^ (new_upper is None), "Bug, please report issue. either both are defined or None."
+#     if new_lower is None:
+#         limits = None
+#     elif new_lower is False:
+#         return False
+#     else:
+#         limits = (new_lower, new_upper)
+#     new_space = zfit.Space(obs=all_obs, limits=limits)
+#     return new_space
```
SebastianJL:
```suggestion
```

zfit/core/integration.py
```diff
@@ -83,69 +83,73 @@ def mc_integrate(func: Callable, limits: ztyping.LimitsType, axes: Optional[ztyp
...
+    integrals = []
+    for space in limits:
+        lower, upper = space._rect_limits_tf
+        tf.debugging.assert_all_finite((lower, upper),
+                                       "MC integration does (currently) not support unbound limits (np.infty) as given here:"
+                                       "\nlower: {}, upper: {}".format(lower, upper)
+                                       )
+
+        # lower = z.convert_to_tensor(lower, dtype=dtype)
+        # upper = z.convert_to_tensor(upper, dtype=dtype)
```
SebastianJL:
```suggestion
```

zfit/core/interfaces.py
```diff
@@ -25,131 +25,253 @@ def copy(self, deep: bool = False, **overwrite_params) -> "ZfitObject":
...
-    @property
-    @abc.abstractmethod
-    def space(self) -> "zfit.Space":
-        """Return the :py:class:`~zfit.Space` object that defines the dimensionality of the object."""
-        raise NotImplementedError
+    # @property
+    # @abstractmethod
+    # def space(self) -> "zfit.Space":
+    #     """Return the :py:class:`~zfit.Space` object that defines the dimensionality of the object."""
+    #     raise NotImplementedError
```
SebastianJL:
```suggestion
```

zfit/core/interfaces.py
```diff
@@ -25,131 +25,253 @@ def copy(self, deep: bool = False, **overwrite_params) -> "ZfitObject":
...
+
+    # TODO: remove below?
+    # @abstractmethod
+    # def get_axes(self, obs: Union[str, Tuple[str, ...]] = None, as_dict: bool = True):
+    #     """Return the axes number of the observable *if available* (set by `axes_by_obs`).
+    #
+    #     Raises:
+    #         AxesNotUnambiguousError: In case
+    #     """
+    #     raise NotImplementedError
```
SebastianJL:
```suggestion
```

zfit/core/interfaces.py
```diff
@@ -25,131 +25,253 @@ def copy(self, deep: bool = False, **overwrite_params) -> "ZfitObject":
...
         raise NotImplementedError
 
-    @abc.abstractmethod
-    def iter_areas(self, rel: bool = False) -> Tuple[float, ...]:
-        """Return the areas of each limit."""
-        raise NotImplementedError
+    # @abstractmethod
+    # def iter_areas(self, rel: bool = False) -> Tuple[float, ...]:
+    #     """Return the areas of each limit."""
+    #     raise NotImplementedError
```
SebastianJL:
```suggestion
```

zfit/core/interfaces.py
```diff
@@ -199,49 +323,102 @@ def with_autofill_axes(self, overwrite: bool):
...
+    #
+    #     Args:
+    #         axes ():
+    #         limits ():
+    #         name (str):
+    #
+    #     Returns:
+    #         :py:class:`~zfit.Space`
+    #     """
+    #     pass
```
SebastianJL: delete

zfit/core/sample.py
```diff
@@ -22,17 +22,28 @@ def __call__(self, n_to_produce: Union[int, tf.Tensor], limits: Space, dtype):
...
+                    ztf.to_real(n_to_produce) * ztf.to_real(frac), dtype=tf.int32)  # TODO(Mayou36): split right!
+
+            # lower = ztf.convert_to_tensor(lower, dtype=dtype)
+            # upper = ztf.convert_to_tensor(upper, dtype=dtype)
+
+            # if isinstance(space, EventSpace):
+            #     lower = tf.transpose(a=lower)
+            #     upper = tf.transpose(a=upper)
+            # if tf.get_static_value(lower.shape[0]) == 1:
+            #     sample_shape = lower.shape[1:]
```
SebastianJL: delete

zfit/core/space.py
```diff
@@ -0,0 +1,2256 @@
+#  Copyright (c) 2020 zfit
```
SebastianJL: Lot's of commented-out lines in this file. Delete them.

zfit/models/basefunctor.py
```diff
@@ -31,20 +32,23 @@ def _check_extract_input_obs(self, obs, models):
...
             obs = models_space
         else:
+            # TODO: why is this here?
+
             if isinstance(obs, Space):
                 obs_str = obs.obs
             else:
                 obs_str = convert_to_container(value=obs, container=tuple)
             # if not frozenset(obs_str) == frozenset(models_space.obs):  # not needed, example projection
             #     raise ValueError("The given obs do not coincide with the obs from the daughter models.")
```
SebastianJL: delete

zfit/models/dist_tfp.py
```diff
@@ -97,11 +97,12 @@ def _unnormalized_pdf(self, x: "zfit.Data", norm_range=False):
...
-        if np.all(-np.array(lower) == np.array(upper)) and np.all(np.array(upper) == np.infty):
-            return z.to_real(1.)  # tfp distributions are normalized to 1
-        lower = z.to_real(lower[0], dtype=self.dtype)
-        upper = z.to_real(upper[0], dtype=self.dtype)
+        lower, upper = limits._rect_limits_tf
+        tf.debugging.assert_all_finite((lower, upper), "Are infinite limits needed? Causes troubles with NaNs")
+        # if np.all(-np.array(lower) == np.array(upper)) and np.all(np.array(upper) == np.infty):
+        #     return z.to_real(1.)  # tfp distributions are normalized to 1
+        # lower = z.to_real(lower[0], dtype=self.dtype)
+        # upper = z.to_real(upper[0], dtype=self.dtype)
```
SebastianJL: delete

zfit/models/physics.py
```diff
@@ -17,7 +17,7 @@ def _powerlaw(x, a, k):
     return a * tf.pow(x, k)
 
 
-@z.function_tf
+@z.function_tf_input
```
SebastianJL: See comment on `zfit/z/zextension.py`

zfit/models/physics.py
```diff
@@ -49,14 +49,16 @@ def crystalball_integral(limits, params, model):
...
+    upper = upper
 
-    return crystalball_integral_func(mu, sigma, alpha, n, lower, upper)
+    integral = crystalball_integral_func(mu, sigma, alpha, n, lower, upper)
+    return integral
 
 
-@z.function_tf
+# @z.function_tf_input
+# @tf.function  # BUG? TODO: problem with tf.function and input signature
```
SebastianJL: delete

zfit/models/physics.py
```diff
@@ -137,15 +141,14 @@ def double_crystalball_mu_integral(limits, params, model):
...
-    upper = upper[0]
+    lower, upper = limits._rect_limits_tf
 
-    #
     return double_crystalball_mu_integral_func(mu=mu, sigma=sigma, alphal=alphal, nl=nl, alphar=alphar, nr=nr,
                                                lower=lower, upper=upper)
 
-@z.function_tf
+
+# @tf.function
```
SebastianJL: delete

zfit/util/cache.py
```diff
@@ -271,11 +275,14 @@ def __eq__(self, other: object) -> bool:
...
         # return all(obj1 == obj2 for obj1, obj2 in zip(self.immutable_representation, other.immutable_representation))
-        import numpy as np
         try:
             return all(np.equal(self.immutable_representation, other.immutable_representation))
         except ValueError:  # broadcasting does not work
             return False
+        except TypeError:  # OperatorNotAllowedError inherits from this
+            return False
+        # except OperatorNotAllowedInGraphError:  # we have to assume they're not the same
+        #     return False
```
SebastianJL: Delete

zfit/util/exception.py
```diff
@@ -1,4 +1,4 @@
-#  Copyright (c) 2019 zfit
+#  Copyright (c) 2020 zfit
```
SebastianJL: Your library uses a lot of extensions. Last time I checked there were even some that are unused. Maybe we should think about reducing them?

zfit/z/zextension.py
```diff
@@ -204,7 +204,7 @@ def concrete_func(*args, **kwargs):
 
 tf_function = FunctionWrapperRegistry()
 
-function_tf = FunctionWrapperRegistry2()  # for only tensorflow inside
+function_tf_input = FunctionWrapperRegistry2()  # for only tensorflow inside
```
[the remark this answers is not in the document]
mayou36: Yes, I agree on the strategy. Actually, the plan is to have a more versatile decorator that can be call with or without arguments.

So tbh, this should not be directly in this PR, and we will address this soon in another branch

tests/test_limit.py
mayou36: > One of the best guidelines I learnt in my internship is the rule of three.

Yes, I agree, just recently learned that too with backends. Either you support 1, 2 or many. One is specific, for two you may still hack. But if you support 3, it's general enough.

> Why do you split this into a function test() and a block of assertions (line 133-139)?

This is because the test is once wrapped within a `tf.function` and once not. This makes it run once _inside a graph context_ and execute (e.g. the 'equal' that is returned is calculated in graph mode).

Some things should hold inside the graph mode, some should be returned correctly (because then they are evaluated).

So the setup, while I agree not great (there is no easy way to tell pytest that it should go also one level deeper to check failed tests? E.g. if an assert inside `test()` fails, it would be good if it points inside the function).

zfit/util/exception.py
mayou36: You mean `exceptions` I assume?

Yeah, it could be reduced maybe. But rather too many, too specific. They are often just to be more informative and with time we figure out better which are needed. But I agree, some unused could be removed

zfit/models/physics.py
mayou36: Maybe I've added the comment afterwards? Should be there since it's an open bug.

zfit/util/cache.py
mayou36: Added a TODO comment

PR #207, opened 2020-03-04
Fix test_get_dependents_ordered() and rename to test_get_dependents_is_deterministic()

SebastianJL:
The pytest test_get_dependents_ordered() tested that the
outcome of list(OrderedDict) != list(set(OrderedDict)).
This is not a guaranteed behavior of the python set() and
list() functions, although it often will be the case.
The actual intention of the test function was to test
whether pdf.get_dependents() would return the same
OrderedDict on each subsequent call, i.e. whether the
function is deterministic. The test function has
been rewritten to reflect that intent.

Fixes #

## Proposed Changes

  -
  -
  -

## Tests added

  -
  -
  -
  
## Checklist

 - [ ] change approved
 - [ ] implementation finished
 - [ ] tests added
 - [ ] CHANGELOG updated

### DIFF
--- a/tests/test_dependents.py
+++ b/tests/test_dependents.py
@@ -6,8 +6,8 @@ import zfit
 from zfit.core.testing import setup_function, teardown_function, tester
 
 
-def test_get_dependents_ordered():
-    params = [zfit.Parameter(f'param{i}', i) for i in range(4)]
+def test_get_dependents_is_deterministic():
+    parameters = [zfit.Parameter(f'param{i}', i) for i in range(4)]
     obs = zfit.Space('obs1', (-3, 2))
 
     def create_pdf(params):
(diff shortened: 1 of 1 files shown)


tests/test_dependents.py
```diff
@@ -6,8 +6,8 @@
 from zfit.core.testing import setup_function, teardown_function, tester
 
 
-def test_get_dependents_ordered():
-    params = [zfit.Parameter(f'param{i}', i) for i in range(4)]
+def test_get_dependents_is_deterministic():
+    parameters = [zfit.Parameter(f'param{i}', i) for i in range(4)]
```
mayou36: Detail: inside we use usually params and param only for the naming, a common abbreviation.

mayou36: but make sense here as well, so perfectly fine actually, sorry!

mayou36:
I'll merge, running just now a few times into this failure

PR #209, opened 2020-03-10
Fix hesse_np and add covariance for any minimizer

marinang:
## Fixes 

* fixes error from `hesse_np` function
* In the `hesse` method, the hessian matrix is now computed using all the free parameters in the fit, leading to the same results between using `hesse_np` and `minuit_hesse` methods when using the `params` argument
* makes `covariance` available for any minimizers

## Tests added

* One test to check that errors from `hesse_np` and `minuit_hesse` are similar.
  
## Checklist

 - [ ] change approved
 - [ ] implementation finished
 - [ ] tests added
 - [ ] CHANGELOG updated

### DIFF
--- a/zfit/minimizers/fitresult.py
+++ b/zfit/minimizers/fitresult.py
@@ -49,9 +49,9 @@ def _hesse_np(result: "FitResult", params, sigma=1.0):
     # if not isinstance(minimizer, Minuit):
     #     raise TypeError("Cannot perform hesse error calculation 'minuit' with a different minimizer then"
     #                     "`Minuit`.")
-    result_hesse = 2 * np.linalg.inv(result.loss.value_gradients_hessian(params)[2])
+    covariance = np.linalg.inv(result.loss.value_gradients_hessian(params)[2])
 
-    hesse = OrderedDict((param, {'error': result_hesse[i, i]}) for i, param in enumerate(params))
+    hesse = OrderedDict((param, {'error': covariance[i, i]**0.5}) for i, param in enumerate(params))
     return hesse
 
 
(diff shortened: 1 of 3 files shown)


zfit/minimizers/fitresult.py
```diff
@@ -40,39 +43,69 @@ def _hesse_np(result: "FitResult", params, sigma=1.0):
...
-    if any([data.weights is not None for data in result.loss.data]):
-        raise WeightsNotImplementedError("Weights are not supported with hesse numpy.")
-
-    # minimizer = fitresult.minimizer
-    # from zfit.minimizers.minimizer_minuit import Minuit
-    # if not isinstance(minimizer, Minuit):
-    #     raise TypeError("Cannot perform hesse error calculation 'minuit' with a different minimizer then"
-    #                     "`Minuit`.")
-    result_hesse = 2 * np.linalg.inv(result.loss.value_gradients_hessian(params)[2])
+    covariance = _covariance_np(result, params)
```
mayou36: What about the weights? Any good idea?
Guess RooFit has a "square the weights, recalculate the likelihood" approximation. Anything better in mind by chance?

mayou36:
You can merge, the CI is fine, the failure is a known bug that is fixed now (I've put it in the wrong branch).

zfit/minimizers/fitresult.py
marinang: Can the hessian matrix can be computed using the weights?

marinang: The procedure is actually described in `Statistical Methods in Experimental Physics` chapter 8.5.2, page 198 if you have it at hand.
Scan_marinang_110320200990_001 - copie

marinang: So as you said squaring the weights and recompute the likelihood.

marinang: But I think you don't have to refit, just clone the loss and square the weights compute the H prime matrix. In that case I think a caching of the covariance matrix should be done in order to avoid the repeat the loss cloning step.

marinang: Here https://arxiv.org/pdf/1911.01303.pdf an asymptotic method is described to compute the covariance matrix.

mayou36: Let's move that discussion to another PR, one thing at a time

PR #210, opened 2020-03-11
Fix typo and add explanation about Functors.

SebastianJL:


mayou36:
I'll merge since right in the middle of a release, good to have this in there as well.

PR #214, opened 2020-03-12
Add errors computed from loss function

### LINKED ISSUE #181: Cannot perform error calculation 'minos_minuit' with a different minimizer then`Minuit`.
I tried to compute the error after fitting with Adam and not Minuit. Apparently this is not possible. 
## Current Behaviour

```# minimize
minimizer = zfit.minimize.Adam()
result = minimizer.minimize(nll)

# calculate errors
param_errors = result.error()
```
The output was: 
```
---------------------------------------------------------------------------

TypeError                                 Traceback (most recent call last)

<ipython-input-17-3f8fdd90f0cd> in <module>()
      3 
      4 # calculate errors
----> 5 param_errors = result.error()

2 frames

/usr/local/lib/python3.6/dist-packages/zfit/minimizers/fitresult.py in _minos_minuit(result, params, sigma)
     48     from zfit.minimizers.minimizer_minuit import Minuit
     49     if not isinstance(minimizer, Minuit):
---> 50         raise TypeError("Cannot perform error calculation 'minos_minuit' with a different minimizer then"
     51                         "`Minuit`.")
     52     result = [minimizer._minuit_minimizer.minos(var=p.name, sigma=sigma)

TypeError: Cannot perform error calculation 'minos_minuit' with a different minimizer then`Minuit`.
```
## Expected Behaviour
I would expect all the three included optimizer to be able to run the error computation.
## Context (Environment)

* zfit version: '0.3.6'
* Python version: '3.6'
* Are you using conda, pipenv, etc? : No, plain python with zfit installed via pip
* Operating System: Debian
* Tensorflow version: 1.14

marinang:
Fixes #181 

## Proposed Changes

  - computation of asymmetric errors from the profiling of the loss function, working with any minimiser
  - this would replace `minos` as the basic function to compute the errors
  - can find new minimum while scanning the loss function, in that case the current fit result attribute `valid` is set to false. The new fit result is returned.

## Tests added

  - check that errors from `minos` and the new function are close/similar
  - check cases when a new minimum is found
  -
  
## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated


zfit/minimizers/errors.py
```diff
@@ -0,0 +1,199 @@
...
+            Computes the pll, with the minimum substracted and shifted by minus the `errordef`, for a
+            given parameter.
+            `errordef` = 1 for a chisquare fit, = 0.5 for a likelihood fit.
+
+            The function raises a `NewMinimum` exception if the value of the shifted `pll` is less than
+            `- errordef`.
+            """
+            if v not in cache:
+                cache[v] = pll(minimizer, loss, param, v) - fmin - errordef
+                if cache[v] < -errordef:
```
mayou36:
```suggestion
                if cache[v] < -errordef - minimizer.tolerance:
```
We don't want to have false positives, right? So a new minimum shouldn't be a fluctuation

mayou36:
Let me know if you think it's ready

marinang:
@mayou36 sorry for being silent (I was preparing my thesis defense :) ). I will add your suggestion but overall I think it’s ready.

zfit/minimizers/errors.py
```diff
@@ -0,0 +1,199 @@
...
+
+    all_params = list(result.params.keys())
+    loss = result.loss
+    errordef = loss.errordef
+    fmin = result.fmin
+    minimizer = result.minimizer.copy()
+    minimizer.tolerance = minimizer.tolerance * 0.5
+    rtol *= errordef
+
+    set_params_to_result(all_params, result)
```
mayou36, line 52:
```suggestion
    set_values(all_params, result)
```
We had the same idea, I've added this functionality independently to `set_values` :)

zfit/minimizers/errors.py
```diff
@@ -0,0 +1,199 @@
...
+        param.floating = True
+    minimizer.verbosity = verbosity
+
+    return minimum.fmin
+
+
+def set_params_to_result(params, result):
+    """Set parameters values to the values in the fitted value in the result."""
+    for param in params:
+        param.set_value(result.params[param]["value"])
```
mayou36, line 35:
```suggestion
```
see below

zfit/minimizers/errors.py
marinang: I yes I saw the new feature,  it will be useful in `hepstats` :D

mayou36:
This looks good so far, just: should we change the default already? Or add it and let people try out? E.g. a "FutureWarning" seems appropriate

marinang:
> This looks good so far, just: should we change the default already? Or add it and let people try out? E.g. a "FutureWarning" seems appropriate

Yes agreed. It's better that way.

PR #227, opened 2020-04-16
Smaller fixes from branch serialization_2020

SebastianJL:
Some smaller fixes and improvements while studying the file parameter.py.

One bigger change is to change ZfitParameter.shape from a property that raises NotImplementedError to an abstract parameter.
# Added Tests
- test_shape_parameter()
- test_shape_composed_parameter()


SebastianJL:
Proposition: Combine BaseComposedParameter and ComposedParameter into one class. What is the benefit of having these two separated?

mayou36:
yeah, that's true actually, I think the idea was to use it as base also for Complex and similar. Let's merge the two PRs though and look again at it later on, okay (otherwise we'll have many conflicts I fear)? maybe it's worth with the dumping to have this extra class and let complex etc inherit from it

PR #228, opened 2020-04-17
Use `value_fn` with args of ComposedParameter

mayou36:
## Proposed Changes

  - value_fn now takes arguments
  -
  -

## Tests added

  - Test with arguments

  
## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -69,7 +69,7 @@ Depreceations
  - `ComposedParameter` takes now `params` instead of `dependents` as argument, it acts now as
    the arguments to the `value_fn`. To stay future compatible, create e.g. `def value_fn(p1, pa2)`
    and using `params = ['param1, param2]`, `value_fn` will then be called as `value_fn(param1, parma2)`.
-   This will most probably break in the future.
+   `value_fn` without arguments will probably break in the future.
 
 
 Bug fixes and small changes
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -67,7 +67,9 @@ Breaking changes
 Depreceations
 -------------
  - `ComposedParameter` takes now `params` instead of `dependents` as argument, it acts now as
-   the arguments to the `value_fn`.
+   the arguments to the `value_fn`. To stay future compatible, create e.g. `def value_fn(p1, pa2)`
+   and using `params = ['param1, param2]`, `value_fn` will then be called as `value_fn(param1, parma2)`.
+   This will most probably break in the future.
 
 
 Bug fixes and small changes
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -64,6 +64,11 @@ Breaking changes
  - SumPDF changed in the behavior. Read above in the Major Features and Improvement.
  - Integrals of extended PDFs are not extended anymore.
 
+Depreceations
+-------------
+ - `ComposedParameter` takes now `params` instead of `dependents` as argument, it acts now as
+   the arguments to the `value_fn`.
+
 
 Bug fixes and small changes
 ---------------------------
--- a/tests/test_parameter.py
+++ b/tests/test_parameter.py
@@ -21,7 +21,7 @@ def test_complex_param():
     def complex_value():
         return real_part + imag_part * 1.j
 
-    param1 = ComplexParameter("param1_compl", complex_value, dependents=None)
+    param1 = ComplexParameter("param1_compl", complex_value, params=None)
     some_value = 3. * param1 ** 2 - 1.2j
     true_value = 3. * complex_value() ** 2 - 1.2j
     assert true_value == pytest.approx(some_value.numpy(), rel=1e-8)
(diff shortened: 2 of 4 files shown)


zfit/core/parameter.py
```diff
@@ -619,6 +622,17 @@ def __init__(self, params, value_fn, name="BaseComposedParameter", **kwargs):
         super().__init__(name=name, params=params, **kwargs)
         if not callable(value_fn):
             raise TypeError("`value_fn` is not callable.")
+        n_params = len(signature(value_fn).parameters)
+        # TODO(0.6): remove, legacy?
+        if n_params == 0:
```
mayou36, line 627: That's the legacy hack, maybe just have a look at these few lines, the rest is fine

tests/test_parameter.py
```diff
@@ -130,11 +134,11 @@ def test_param_limits():
 
 
 def test_overloaded_operators():
-    param_a = ComposedParameter('param_a', lambda: 5 * 4, dependents=None)
-    param_b = ComposedParameter('param_b', lambda: 3, dependents=None)
+    param_a = ComposedParameter('param_a', lambda: 5 * 4, params=None)
```
SebastianJL, line 137: Won't this make problems with the check in BaseComposedParameter for the signature of the value_fn, i.e. I think this issues a Warning. Or is this just for legacy purposes?

zfit/core/parameter.py
SebastianJL: Maybe give a proper deprecation warning?

SebastianJL: And you are not testing for what you are claiming to test in the warning.

CHANGELOG.rst
```diff
@@ -64,6 +64,11 @@ Breaking changes
  - SumPDF changed in the behavior. Read above in the Major Features and Improvement.
  - Integrals of extended PDFs are not extended anymore.
 
+Depreceations
+-------------
+ - `ComposedParameter` takes now `params` instead of `dependents` as argument, it acts now as
+   the arguments to the `value_fn`.
+
```
SebastianJL, line 71: I think we should state more explicitly that value_fn now needs to take explicit arguments instead of referencing the locals() and globals() dictionaries and implicate that this will force the user to change their value_fn if they want to stay compatible in the future.

zfit/core/parameter.py
mayou36: Yes, true. I've added the real test and changed to a depr warning

mayou36:
Also updated the docs, yep

CHANGELOG.rst
```diff
@@ -67,7 +67,9 @@ Breaking changes
 Depreceations
 -------------
  - `ComposedParameter` takes now `params` instead of `dependents` as argument, it acts now as
-   the arguments to the `value_fn`.
+   the arguments to the `value_fn`. To stay future compatible, create e.g. `def value_fn(p1, pa2)`
+   and using `params = ['param1, param2]`, `value_fn` will then be called as `value_fn(param1, parma2)`.
+   This will most probably break in the future.
```
SebastianJL, line 72:
```suggestion
   `value_fn` without arguments will probably break in the future.
```

zfit/core/parameter.py
```diff
@@ -623,15 +623,19 @@ def __init__(self, params, value_fn, name="BaseComposedParameter", **kwargs):
...
         if n_params == 0:
-            warnings.warn("The `value_fn` for composed parameters should take the same number"
-                          " of arguments as `params` are given.")
-            legacy_value_fn = value_fn
-
-            def value_fn(*_):
-                return legacy_value_fn()
-        # end legacy
+            if len(params) == 0:
+                warnings.warn("No `params` specified, the `value_fn` is supposed to return a constant. "
```
SebastianJL, line 629: This warning might also make sense if `n_params > 0`. Maybe add a test to see if `n_params == len(params)`.
Also naming is a bit confusing here due to the double meaning of params. Maybe call it `n_function_parameters` instead of `n_params`.

tests/test_parameter.py
```diff
@@ -74,19 +74,23 @@ def test_composed_param():
     param1 = Parameter('param1', 1.)
     param2 = Parameter('param2', 2.)
     param3 = Parameter('param3', 3., floating=False)
-    param4 = Parameter('param4', 4.)  # needed to make sure it does not only take all params as deps
+    param4 = Parameter('param4', 4.)  # noqa Needed to make sure it does not only take all params as deps
```
SebastianJL, line 77:
```suggestion
    param4 = Parameter('param4', 4.)  # Needed to make sure it does not only take all params as deps
```

zfit/core/parameter.py
mayou36: I did not add the test on purpose, as there can be variable arguments lengths in the signature. Renaming `n_params` makes sense

tests/test_parameter.py
mayou36: noqa should be there, otherwise it's checked by linters as an unused parameter

PR #241, opened 2020-05-28
New algorithm to compute PLL intervals

marinang:
## Proposed Changes

Computes the PLL intervals using the algorithm described in Venzon and Moolgavkar in *A Method for Computing Profile-Likelihood-Based Confidence Intervals*, Journal of the Royal Statistical Society C37 (1988) 87–94 [DOI](https://www.jstor.org/stable/2347496) , but using a root-finding algorithm from SciPy.

## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated

### DIFF
--- a/tests/test_fitresult.py
+++ b/tests/test_fitresult.py
@@ -170,17 +170,18 @@ def test_correlation(minimizer_class_and_kwargs):
 
 
 @pytest.mark.parametrize("minimizer_class_and_kwargs", minimizers)
-def test_errors(minimizer_class_and_kwargs):
+@pytest.mark.parametrize("sigma", [1, 2])
+def test_errors(minimizer_class_and_kwargs, sigma):
     results = create_fitresult(minimizer_class_and_kwargs=minimizer_class_and_kwargs)
     result = results['result']
     a = results['a_param']
     b = results['b_param']
     c = results['c_param']
 
-    z_errors, new_result = result.errors(method="zfit_error")
+    z_errors, new_result = result.errors(method="zfit_error", sigma=sigma)
     if new_result is not None:
         result = new_result
-    minos_errors, _ = result.errors(method="minuit_minos")
+    minos_errors, _ = result.errors(method="minuit_minos", sigma=sigma)
 
     for param in [a, b, c]:
         z_error_param = z_errors[param]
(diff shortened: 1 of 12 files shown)


tests/test_fitresult.py
```diff
@@ -49,6 +51,9 @@ def create_fitresult(minimizer_class_and_kwargs):
     minimizer = minimizer_class(**minimizer_kwargs)
 
     result = minimizer.minimize(loss=loss)
+    print(result)
+    result.status
+    result.info
```
mayou36, line 56:
```suggestion
```
Remove that I assume?

zfit/minimizers/errors.py
```diff
@@ -174,19 +40,58 @@ def compute_errors(result, params, sigma=1, rootf=_rootf, rtol=0.01):
...
+            param_error = covariance[(param, param)]**0.5
+            param_value = result.params[param]["value"]
+            other_params = [p for p in all_params if p != param]
+
+            initial_values = {"lower": [], "upper": []}
+            direction = {"lower": -sigma, "upper": sigma}
+
+            for ap in all_params:
+                for d in ["lower", "upper"]:
+                    # shift parameters, other than param, using covariance matrix
```
mayou36, line 66: Smart idea!

tests/test_fitresult.py
marinang: Haha yes It can be removed if you are not nervous about coverage score :P. Indeed printing and those attributes are never checked and thus lowers the coverage of fit_result.py.

mayou36: :D I see, so I would rather put it somewhere else then and do a check actually. But yeah, pushing the coverage would be nice actually

marinang: Yeah I agree with that.

tests/test_fitresult.py
```diff
@@ -163,62 +168,75 @@ def test_correlation(minimizer_class_and_kwargs):
...
     a = results['a_param']
     b = results['b_param']
     c = results['c_param']
 
-    z_errors, new_result = result.errors(method="zfit_error")
-    if new_result is not None:
-        result = new_result
-    minos_errors, _ = result.errors(method="minuit_minos")
-
+    for n_trial in range(n_max_trials):
```
mayou36, line 182: I've added a loop that retries with the new result if a new minimum was found, otherwise the test would be "silently skipped". If you're ok with that, you can merge, the rest looks good!

PR #244, opened 2020-06-08
computing covariance/errors with weighted events

marinang:
## Proposed Changes

Implements the correction to the covariance matrix described in this [paper](https://arxiv.org/pdf/1911.01303.pdf) in equation 18, for weighted events (without uncertainties). 

## Tests added

  - add test with weights in test_covariance
  
## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated


mayou36:
Looks good, I fixed the conflict. Let's merge and we can refactor things out later, let's put it into testing mode with the public ;)

marinang:
@mayou36 yep agreed :)

mayou36:
Okey, we don't have a test yet, right, or did you not push it? Otherwise I'll add something later

marinang:
Right I didn't, I can also design a test later today.

mayou36 (approved):

marinang:
Maybe we should add a warning message to say that this is experimental?

PR #245, opened 2020-06-09
add 1dim Gaussian KDE

mayou36:
## Proposed Changes

  - add a 1dim KDE PDF

## Tests added

  - test_kde_pdf
  
## Checklist

 - [ ] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated

### DIFF
--- a/zfit/models/kde.py
+++ b/zfit/models/kde.py
@@ -6,11 +6,11 @@ import tensorflow as tf
 import tensorflow_probability as tfp
 from tensorflow_probability.python import distributions as tfd
 
-from ..core.interfaces import ZfitData
 from .dist_tfp import WrapDistribution
+from .. import z, ztypes
+from ..core.interfaces import ZfitData
 from ..util import ztyping
 from ..util.exception import OverdefinedError, ShapeIncompatibleError
-from .. import z, ztypes
 
 
 def bandwidth_rule_of_thumb(data, factor=0.9):
@@ -60,7 +60,11 @@ class GaussianKDE1DimExactV1(WrapDistribution):
     def __init__(self, obs: ztyping.ObsTypeInput, data: ztyping.ParamTypeInput,
                  bandwidth: ztyping.ParamTypeInput = None,
                  weights: Union[None, np.ndarray, tf.Tensor] = None, name: str = "GaussianKDE1DimV1"):
-        """One dimensional Kernel Density Estimation with a Gaussian Kernel.
+        r"""One dimensional Kernel Density Estimation with a Gaussian Kernel.
+
+        .. math::
+
+            f_h(x) =  \frac{1}{nh} \sum_{i=1}^n K\Big(\frac{x-x_i}{h}\Big)
 
         The bandwidth of the kernel can be estimated in different ways. It can either be a global bandwidth,
         corresponding to a single value, or a local bandwidth
--- a/zfit/models/kde.py
+++ b/zfit/models/kde.py
@@ -8,7 +8,7 @@ from tensorflow_probability.python import distributions as tfd
 
 from .dist_tfp import WrapDistribution
 from .. import z, ztypes
-from ..core.interfaces import ZfitData
+from ..core.interfaces import ZfitData, ZfitSpace
 from ..util import ztyping
 from ..util.exception import OverdefinedError, ShapeIncompatibleError
 
@@ -34,9 +34,9 @@ def bandwidth_adaptiveV1(data, func):
     return bandwidth
 
 
-def _adaptive_bandwidth_KDEV1(constructor, obs, data, weights, name):
+def _adaptive_bandwidth_KDEV1(constructor, obs, data, weights, name, truncate):
     kde_silverman = constructor(obs=obs, data=data, bandwidth='silverman', weights=weights,
-                                name=f"INTERNAL_{name}")
+                                name=f"INTERNAL_{name}", truncate=truncate)
     return bandwidth_adaptiveV1(data=data, func=kde_silverman.pdf)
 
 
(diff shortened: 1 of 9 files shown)


zfit/models/kde.py
```diff
@@ -0,0 +1,134 @@
...
+
+from .dist_tfp import WrapDistribution
+from .. import z, ztypes
+from ..core.interfaces import ZfitData
+from ..util import ztyping
+from ..util.exception import OverdefinedError, ShapeIncompatibleError
+
+
+def bandwidth_rule_of_thumb(data, factor=0.9):
+    return tf.math.reduce_std(data) * tf.cast(tf.shape(data)[0], ztypes.float) ** (-1 / 5.) * factor
```
marinang, line 17: Shouldn't the IQR be used as well as it is more robust for long-tailed and skew distributions.

```
return tf.minimum(tf.math.reduce_std(data), (tfp.stats.percentile(data, 75) - tfp.stats.percentile(data, 25)) / 1.34) * tf.cast(tf.shape(data)[0], ztypes.float) ** (-1 / 5.) * factor
```

zfit/models/kde.py
```diff
@@ -0,0 +1,134 @@
...
+def bandwidth_scott(data):
+    return bandwidth_rule_of_thumb(data=data, factor=1.059)
+
+
+def bandwidth_adaptiveV1(data, func):
+    from .. import run
+    run.assert_executing_eagerly()
+    data = z.convert_to_tensor(data)
+    bandwidth = z.sqrt(tf.math.reduce_std(data) / func(data))
+    bandwidth *= tf.cast(tf.shape(data)[0], ztypes.float) ** (-1 / 5.) * 1.059
```
marinang, line 33: Just a curiosity what kind of adaptive bandwidth did you implement here?. I see many things in the literature but the most common on is $$h(x) = \frac{h_0}{pdf(x)^2 \times G}  $$ where $$pdf$$ is the first estimate of density as you did and $$G$$ is the geometric mean of the data.

zfit/models/kde.py
mayou36: Yeah, it seems like. Do you refer to the one on [Wikipedia](https://en.wikipedia.org/wiki/Kernel_density_estimation)?

marinang: Yes indeed.

zfit/models/kde.py
mayou36: I took eq. 6 from [Kyles Kernel Density in HEP](https://reader.elsevier.com/reader/sd/pii/S0010465500002435?token=BCE1BEFF1924EF76532CCF9D2D3F0274B4C41E49DB20B0CBFEA2A32AE5FD51826548F615DCCEA29D5B4AA65EF524DFD0) (modulo using also the IQR, should put that there). However, that square root and why "it clearly follows" is (intuitively) not fully clear to me.

Where did you find the geometric mean? I would not quite understand why we would need that for the bandwidth (the spread)?

marinang: This is how it is done in [Stata](https://www.stata.com/meeting/9uk/akdensity.pdf) but this was also described in a Silverman's book (searching for a link).

marinang: > I took eq. 6 from [Kyles Kernel Density in HEP](https://reader.elsevier.com/reader/sd/pii/S0010465500002435?token=BCE1BEFF1924EF76532CCF9D2D3F0274B4C41E49DB20B0CBFEA2A32AE5FD51826548F615DCCEA29D5B4AA65EF524DFD0) (modulo using also the IQR, should put that there). However, that square root and why "it clearly follows" is (intuitively) not fully clear to me.
> 
> Where did you find the geometric mean? I would not quite understand why we would need that for the bandwidth (the spread)?
 
You mean that `h* ∝ σ **2` ?. This I succeeded to convince me about that since `σ ∝ n**0.5`, but still I don't get how Eq. 6 is obtained.

marinang: Anyway as long as new bandwidth scales with the inverse of the pilot pdf scare it's fine, then it's juste fine tuning. The reason for the geometric mean is explained in Silverman .
Capture d’écran 2020-06-11 à 12 09 20


Capture d’écran 2020-06-11 à 12 08 59

mayou36: I see, G refers to the geometric mean of the pilot estimate, e.g. of f_0(x_i), not of the data points (so kinda the average density response of our first estimate)

mayou36: So here two things are used, lambda and h, currently we place them together. It could surely be cleaner though to have both, I agree

marinang: Ah right I said the geometric mean on the data, but it's indeed on the pilot estimate.  Yeah having both h and λ = 1 by default for fixed kernel bandwidth, and then λ = something / √𝓅𝒹𝒻 for adaptive.

zfit/models/kde.py
mayou36: I've added that

zfit/models/kde.py
mayou36: While I agree on the nice splitting into h and lambda, it seems a little bit more complicated to generalize properly with parameters, as there are currently only single valued parameters and how to handle it. My idea is to go with the simpler layout (currently) of having just a bandwidth that can be specified. While this is somewhat limited, we can get some experience with it and then better generalize (also to ndims). Sounds reasonable?

zfit/models/kde.py
```diff
@@ -0,0 +1,166 @@
...
+def _bandwidth_scott_KDEV1(data, *_, **__):
+    return bandwidth_scott(data)
+
+
+def _bandwidth_silverman_KDEV1(data, *_, **__):
+    return bandwidth_silverman(data)
+
+
+@z.function(wraps='tensor')
+def min_std_or_iqr(x):
```
marinang, line 52: This doesn't seem to be used anywhere.

zfit/models/kde.py
marinang: Yes it's reasonable, improvements can come later.

zfit/models/kde.py
mayou36: fixed, thanks!

PR #252, opened 2020-07-03
Fix GaussianConstraint example

olantwin:
## Proposed Changes

  - Fixes the example for GaussianConstraint in the documentation as the keyword arguments were renamed in the constructor.

## Checklist

 - [ ] change approved
 - [ ] implementation finished
 - [ ] tests added
 - [ ] CHANGELOG updated


mayou36:
Thanks for spotting!

PR #258, opened 2020-07-14
Remove type hints in docstrings.

SebastianJL:
Fixes:

- Duplication between docstrings and type annotations.
- sphinx_autodoc_typehints not being able to render the type annotations.

## Proposed Changes

  - Delete all type hints inside the docstrings.

## Tests added

  - None
  
## Checklist

 - [ ] change approved
 - [ ] implementation finished
 - [ ] tests added
 - [ ] CHANGELOG updated

### DIFF
--- a/zfit/z/zextension.py
+++ b/zfit/z/zextension.py
@@ -38,9 +38,9 @@ def nth_pow(x, n, name=None):
     """Calculate the nth power of the complex Tensor x.
 
     Args:
-        x (tf.Tensor, complex):
-        n (int >= 0): Power
-        name (str): No effect, for API compatibility with tf.pow
+        x:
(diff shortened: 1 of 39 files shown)


zfit/z/zextension.py
```diff
@@ -38,9 +38,9 @@ def nth_pow(x, n, name=None):
     """Calculate the nth power of the complex Tensor x.
 
     Args:
-        x (tf.Tensor, complex):
-        n (int >= 0): Power
-        name (str): No effect, for API compatibility with tf.pow
+        x:
+        n: Power
```
mayou36, line 42:
```suggestion
        n: Power of x, has to be a positive int
```

PR #264, opened 2020-09-30
Addition of Poisson 

marinang:
## Proposed Changes

  - Add the Poisson PDF, this will be useful for analysts who want to do inferences with `hepstats`  with counting analysis instead of shape analysis (see https://github.com/scikit-hep/hepstats/pull/33). And maybe for other cases ... 🤷🏻‍♂️

## Tests added

  - test_pdf_poisson.py
  
## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -35,6 +35,7 @@ Thanks
 Major Features and Improvements
 -------------------------------
 - completely new doc design
+- new Poisson PDF
 
 Breaking changes
 ------------------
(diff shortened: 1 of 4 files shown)


mayou36:
Good idea! Can you import it to the correct namespace, then we can merge

marinang:
> Good idea! Can you import it to the correct namespace, then we can merge

I am not sure what is incorrect here?

mayou36 (approved):
Yep, that's what I meant. Perfect!

PR #270, opened 2020-10-09
WIP Binned

mayou36:
Fixes #

## Proposed Changes

  -

## Tests added

  -
  
## Checklist

 - [ ] change approved
 - [ ] implementation finished
 - [ ] tests added
 - [ ] CHANGELOG updated

### DIFF
--- a/zfit/core/loss.py
+++ b/zfit/core/loss.py
@@ -7,15 +7,18 @@ import tensorflow as tf
 from typing import Optional, Union, List, Callable
 
 from zfit import ztf
-from zfit.util import ztyping
-from zfit.util.cache import Cachable
-from zfit.util.graph import get_dependents_auto
+from zfit.core.data import HistData
+from .histogram import midpoints_from_hist
+from ..util import ztyping
+from ..util.cache import Cachable
+from ..util.graph import get_dependents_auto
+from ..util.temporary import TemporarilySet
 from .baseobject import BaseObject, BaseDependentsMixin
 from .interfaces import ZfitLoss, ZfitSpace, ZfitModel, ZfitData, ZfitPDF
 from ..models.functions import SimpleFunc
 from ..util.container import convert_to_container, is_container
-from ..util.exception import IntentionNotUnambiguousError, NotExtendedPDFError
-from zfit.settings import ztypes
+from ..util.exception import IntentionNotUnambiguousError, NotExtendedPDFError, DueToLazynessNotImplementedError
+from ..settings import ztypes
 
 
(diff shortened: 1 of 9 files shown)


zfit/core/loss.py
```diff
@@ -246,6 +284,55 @@ def _gradients(self, params):
         return param_gradients
 
 
+def hist_to_probs_midpoints(data: HistData, model):
+    with data.set_value_converter(midpoints_from_hist):
+        probs = model.pdf(data)
```
marinang, line 289: I think I saw several times Hans writing about binned likelihood, and that they should be computed with the CDF instead of the CDF.

marinang:
Nice work, I had a look at the PR by curiosity :). BTW there are two PR with the same name.

zfit/core/loss.py
mayou36: Yes, he also uses this in iminuit. However, the first thing that is done in the cost is to convert it to a pdf with np.diff (or am I mistaken?): https://github.com/scikit-hep/iminuit/blob/21e14425cdd491086d94b3a262bbd1fa2ec6c93e/src/iminuit/cost.py#L201

mayou36:
> Nice work, I had a look at the PR by curiosity :). BTW there are two PRs with the same name.

Yes, in fact both of them are not the real ones, the `binned_new` is the actual binned one. We had quite a few discussions again and I guess we finally arrived at a good design.

zfit/core/loss.py
marinang: It is not a conversion to a pdf but to a probability.

mayou36: Ah, being sloppy here with the language or pdf <-> probability: I agree. I meant to use the "binned PDF" (=probabilities), which is what Hans uses as well.
Anyway, the real implementation is in the `binned_new`.

The point about CDF vs PDF: they are equivalent _if you integrate the pdf over the bin_. Meaning if we have both available, the CDF and the PDF analytically, it doesn't matter which one you take (just if you're providing a small function as in iminuit case, you'd rather ask for the CDF).

However, in the general case, we can better approximate an integral inside a bin instead of a CDF (in the former, we need to integrate effectively `n_bins` times while in the latter we need "infinite" many integrals)

If you're available for a chat anytime, I'd be happy to and see what you do think about the design

marinang: Sure Indeed the CDF is just the integral of the PDF. Actually this is something I do in `hepstats` where generating the [asimov histogram](https://github.com/scikit-hep/hepstats/blob/17ed8d7818de275bb04af011ec364b2a171966cc/src/hepstats/hypotests/calculators/asymptotic_calculator.py#L12).

PR #285, opened 2021-02-12
Incorporating new 1D KDE PDFs

AstroViking:
## Proposed Changes

  - Implement three new one-dimensional kernel density estimation PDFs, namely KDE1DimV1, KDE1DimFFTV1, KDE1DimISJV1

## Tests added

  -
  
## Checklist

 - [x] change approved
 - [x] implementation finished
 - [x] tests added
 - [x] CHANGELOG updated


mayou36:
Hey, sorry for not writing here, I've been thinking about how and where to put things, including how to name stuff and also in the docs.

What I thought could be nice is to have a dedicated docs page on KDEs actually, where we can put, some text about the KDEs (basically taken from your thesis) and also link to it. What do you think?

The tests can basically be copied (or better, parametrized) form the current KDE one, maybe I can actually do that.

mayou36:
just FYI, this got delayed because it has a dependency on the tf financial which is py37+ and we still have 3.6 as well. But planned to drop it next TF release ~ beginning of May

jonas-eschle:
Hey @AstroViking , I've added some stuff to the implementation (basically to have an "exactKDE" and a "GridKDE", added some mirroring of the data as an option, defaults, a whole array of tests and stuff... It works now well I think and is mainly missing the documentation.

There should surely be some docstring in the method, I am taking care of that, but we should have also:
- a reference to your thesis (I am using now the website of your repo)
- a more in-depth explanation (which could be parts of the Thesis copied at the right place)

Would you want to write a bit an explanation of things, or are you fine if I copy some parts of the thesis?

jonas-eschle:
@AstroViking I'll merge this, in case you disagree with the docs and the citing or anything like that, just let me know. Thanks again for all your great work!

PR #288, opened 2021-02-23
Fixing Flaky test

sleepy-owl:
The test `test_onedim_sampling` is flaky. This PR fixes this issue.

To find a solution, I collected samples of the pvalue from several test executions and computed the tail distribution. I computed the extreme percentiles to check how high can the values be.

```
0.9::1.5e-05
0.99::1.22e-07
0.9999::2.29e-8 ~ 1e-8
```

For this fix, I chose the 99.99th percentile. I think setting the bound using the statistical evaluation might be a good way to ensure the test is not flaky.

Do you guys think this makes sense? Please let me know if this looks good or if you have any other suggestions. Also, here I assume there are no bugs in the code under test.

### DIFF
--- a/tests/test_pdf_fftconv.py
+++ b/tests/test_pdf_fftconv.py
@@ -121,7 +121,7 @@ def test_onedim_sampling():
     sample_nosample = conv_nosample.sample(npoints_sample)
     x = z.unstack_x(sample)
     xns = z.unstack_x(sample_nosample)
-    assert scipy.stats.ks_2samp(x, xns).pvalue > 1e-8  # can vary a lot, but still means close
+    assert scipy.stats.ks_2samp(x, xns).pvalue > 1e-7  # can vary a lot, but still means close
     # uni1sample = func1k.sample(npoints_sample // 2)
     # uni2sample = func2k.sample(npoints_sample // 2)
     # xcomb = tf.concat([uni1sample.value(), uni2sample.value()], axis=0)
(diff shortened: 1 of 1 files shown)


mayou36:
Hey, many thanks!, Why did you choose exactly this test? Just because there are many that are potentially flaky

sleepy-owl:
Actually, I ran some of your tests 500 times to check for flakiness. I observed 2 tests failed once or more. This was one of them. I also have a fix for the other test. I can raise another PR for that if you would like that.

Obviously, I might have missed other flaky tests in your project. Can you point me to some other flaky tests that you know of? I can try fixing them as well!

Also, let me know if this fix makes sense?

Thanks!

mayou36 (approved):

mayou36:
I've put it to 1e-7, it is actually already way too low and we should think of another way of handling it...

Just out of curiosity, how did it come that you've been doing that?

sleepy-owl:
Oh, do you mean why I am fixing these tests? It is part of a research project that I have doing recently! I would be happy to learn about any other potential flaky tests that you know of. Thanks!

mayou36:
Oh, that is interesting! What is it about then?

For the test, I have to check, but will let you know!

sleepy-owl:
It is about fixing these kind of flaky tests. So I am on the lookout for more :) 

If this change seems reasonable, can you please merge it? Much appreciated!

PR #289, opened 2021-02-23
Update 5_minutes_to_zfit.rst

donalrinho:
Added additional information for the use of Hesse instead of Minos to derive parameter errors. Important for users dealing with weighted data, where Minos should not be used.

Fixes #

## Proposed Changes

  -

## Tests added

  -
  
## Checklist

 - [ ] change approved
 - [ ] implementation finished
 - [ ] tests added
 - [ ] CHANGELOG updated


mayou36:
I've added it before the errors actually, as it is in general easier to use, if applicable.

mayou36 (approved):

PR #290, opened 2021-02-25
Fixing test flakiness

sleepy-owl:
The test `test_sampling` is flaky. It sometimes fails when the p-value is below 0.05. i collected the samples of p-value from several executions and computed the tail distribution. The computed percentiles are as follows:

```
0.9 :: 0.096
0.99 :: 0.018
0.9999 :: 0.0114
```

I updated the assertion using the 99.99th percentile. Do you think looks reasonable? I will be happy to incorporate any other suggestions that you may have.

Thanks!

### DIFF
--- a/tests/test_pdf_sumpdf.py
+++ b/tests/test_pdf_sumpdf.py
@@ -94,7 +94,7 @@ def test_sampling():
     assert true_mu == pytest.approx(np.mean(sample), abs=tolerance)
     assert np.std(sample_true) == pytest.approx(np.std(sample), abs=tolerance)
 
-    assert scipy.stats.ks_2samp(sample_true, sample).pvalue > 0.05
+    assert scipy.stats.ks_2samp(sample_true, sample).pvalue > 0.01
 
 
 @pytest.mark.flaky(2)  # mc integration


mayou36:
How often does it fail then?

sleepy-owl:
The original version of the test failed 1/500 times for me. Also, since it marked as flaky, I see that it was re-run at least once for about 50 times. 

Do you want to know how often it fails after the fix?

sleepy-owl:
I ran the test 1000 times after fixing, it failed 0 times and was re-run 14 times

mayou36:
I see, there is however a problem to that: we now know that it won't fail if it is correct. But do we also know that it fails if it is wrong?

That's the tricky part here: How do we make sure that we don't just make it pass all the time?

Do you have an idea?

sleepy-owl:
That is an interesting question. For my analysis, i used statistical tests to determine the range of values that can occur during execution and determine appropriate bounds using that. Here my assumption is that the current version of the code has no bugs. Hence, if a bug causes the values to go beyond the expected range, this test should still catch it. 

In the context of catching such accuracy regressions, do you know if the original version of the test is effective in catching such bugs? Are there any instances from previous builds or from your personal experience where this test (or a similar test) detected a bug introduced by a change? If yes, we could probably check if such a change would also trigger a failure in this fixed version of the test and give us some confidence in this version.

Let me know what you think!

mayou36:
> In the context of catching such accuracy regressions, do you know if the original version of the test is effective in catching such bugs? Are there any instances from previous builds or from your personal experience where this test (or a similar test) detected a bug introduced by a change? If yes, we could probably check if such a change would also trigger a failure in this fixed version of the test and give us some confidence in this version.

well, it's all probability: we could optimize the power, but I guess for this case it seems fine enough for me.

Does it make sense? alpha and beta rates...

Gonna close this otherwise

sleepy-owl:
Since the test is already marked flaky, i guess then there is not much to do here other than increasing number of re-runs (if you dont want to update the assertion). Running 3 times should bring down the chances of failure to ~0.01%.

mayou36:
So I guess it is fine as it is, but many thanks for the suggestions

PR #310, opened 2021-04-01
Tensorflow numpy api

SebastianJL:
## Proposed Changes

- Replace part of the tensorflow namespace with the `tensorflow.experimental.numpy` namespace.
- **Update to tensorflow~=2.5.0rc0**
- Change behaviour of `ComplexParameter` properties `imag`, `real`, `arg`, `mod`. They now always return a `tf.Tensor` and not a `Parameter`.

### Replaced functions
Roughly two types of functions were changed:
- Functions from the `tensorflow.math` namespace.
- Functions that construct new Tensors based on existing tensors like `tf.concat`, `tf.stack` or `tf.transpose`.

### Skipped functions
- All functions that create a `tf.Tensor` from scratch like `tf.constant`, `tf.variable`, `tf.range` or `tf.convert_to_tensor`.
- Functions that don't have an equivalent in tensorflow NumPy. This are mainly one of the following:
    + Function exists in numpy but has not yet been implemented in tensorflow NumPy.
    + Function does not exist in numpy but would make sense to be there. :wink: (e.g. `tf.unstack`).
    + Everything under `tf.debbuging.assert_*`.
    + Flow control functions like `while_loop`.
    + Logical functions like `tf.greater`.

### Skipped regions
- Everything under `test`, as to not change what is actually tested. This does not mean, that there are no changes to the tests but that there are no replacements of tensorflow functions with `tensorflow.experimental.numpy` functions.
- Examples, docs and docstrings, e.g. everything that appears to the user on the website.

## Notes

- The modules `z.unstable`, `z.wrapping_tf` and `z.zextension` now both use tensorflow NumPy and pure tensorflow functions in an inconsistent manner. The purpose of these three modules seems overlapping at times and I think it would make sense to harmonize them at a later point when tensorflow NumPy has proven to be stable enough. I.e. remove parts of it in favour of tensorflow numpy and consolidate the parts that are still necessary. This is also connected to the next point.
- `tf.experimental.numpy.experimental_enable_numpy_behavior()` is not called yet. It does the following things
    + Enables type promotion in TensorFlow
    + Changes type inference, when converting literals to tensors, to more strictly follow the NumPy standard.
    + Adds methods and properties to `tf.Tensor` like `astype()` and `T` to make it compatible with the numpy ndarray. This last behaviour is not documented yet, as it seems to be new in `tensorflow~=2.5.0rc0`. This is probably due to the decision to scrap the separate `tf.experimental.numpy.ndarray` class and just make it an alias of `tf.tensor` (See commit `Add test for zfit/z/numpy.py.` 7656cba7faf6ff74cfa5701e993514b28fb3dc8d)

## Tests added

- Tests for checking that `ComplexParamter.imag/real/arg/mod` always return a `tf.Tensor`.

## Checklist

-   [ ] change approved
-   [ ] implementation finished
-   [ ] correct namespace imported
-   [ ] tests added
-   [ ] CHANGELOG updated

### DIFF
--- a/tests/test_z_numpy.py
+++ b/tests/test_z_numpy.py
@@ -1,0 +1,17 @@
+#  Copyright (c) 2021 zfit
+import tensorflow as tf
+
+import zfit.z.numpy as znp
+
+
+def test_z_numpy_ndarray_is_tensorflow_tensor():
+    """In tensorflow 2.4.1 tf.experimental.numpy.ndarray was a wrapper around tf.Tensor.
+
+    Now this concept seems to
+    have been scratched and tf.experimental.numpy.ndarray is just an alias for tf.Tensor.
+    See the commit history of
+    https://github.com/tensorflow/tensorflow/commits/master/tensorflow/python/ops/numpy_ops/np_arrays.py
+    """
+    assert znp.ndarray is tf.Tensor
+    assert isinstance(znp.array(1), tf.Tensor)
+    assert isinstance(znp.sum(znp.array(0)), tf.Tensor)
--- a/zfit/models/functor.py
+++ b/zfit/models/functor.py
@@ -13,6 +13,7 @@ from typing import List, Optional
 import tensorflow as tf
 
 import zfit.z.numpy as znp
+
 from .. import z
 from ..core.basepdf import BasePDF
 from ..core.coordinates import convert_to_obs_str
--- a/zfit/models/functor.py
+++ b/zfit/models/functor.py
@@ -12,6 +12,7 @@ from typing import List, Optional
 
 import tensorflow as tf
 
+import zfit.z.numpy as znp
 from .. import z
 from ..core.basepdf import BasePDF
 from ..core.coordinates import convert_to_obs_str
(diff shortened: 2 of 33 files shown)


SebastianJL:
@mayou36 I am finished with this PR. There are still problems in the checks, but I don't see any that are related to this PR and we might not want to fix them here.
Have a look at the PR and then we can discuss any further questions or remarks you might have.

tests/test_z_numpy.py
```diff
@@ -0,0 +1,17 @@
+#  Copyright (c) 2021 zfit
+import tensorflow as tf
+
+import zfit.z.numpy as znp
+
+
+def test_z_numpy_ndarray_is_tensorflow_tensor():
+    """In tensorflow 2.4.1 tf.experimental.numpy.ndarray was a wrapper around tf.Tensor.
```
mayou36, line 8: Nice check!

zfit/models/functor.py
```diff
@@ -203,8 +203,7 @@ def _pdf(self, x, norm_range):  # NOT _pdf, as the normalization range can diffe
         pdfs = self.pdfs
         fracs = self.params.values()
         probs = [pdf.pdf(x) * frac for pdf, frac in zip(pdfs, fracs)]
-        prob = functools.reduce(operator.add, probs)
-        # prob = tf.math.accumulate_n([pdf.pdf(x) * frac for pdf, frac in zip(pdfs, fracs)])
+        prob = sum(probs)
```
mayou36, line 206: :D I missed here the obvious solution, good catch

mayou36 (approved):
Alright, that looks all good to me. This is not WIP anymore I assume, right? Feel free to merge

PR #335, opened 2021-05-11
Fix typos in the intro guide

simonthor:
# Fixes
While going through the documentation, specifically the intro section, I noticed some minor typos. I probably missed some but here are the ones that I found and fixed.


mayou36:
Good catches, many thanks!

PR #339, opened 2021-05-15
WIP: reshuffle docs

### LINKED ISSUE #325: Shifting the upcoming tutorials to a new space in the documentation
The newly added/upcoming tutorials should be moved to a completely different section from the getting started section. I believe that as more tutorials will be added, the complexity of the same will increase and thus they will not fit in the current section which focuses more on the introductory concepts. 

By adding a new section:
- The users will be able to access it more efficiently as the new section would increase its visibility.
- Categorization of tutorials could be done more efficiently in the future on the basis of the usage of advanced concepts.

mayou36:
Fixes #325 

## Proposed Changes

- move the docs around, see in live preview here: https://zfit--339.org.readthedocs.build/en/339/whats_new/index.html

It mainly consists of new tutorial ordering:
- introduction: getting people started
- components: explaining parts of zfit like data, model etc
- statistics with zfit: to come, more focused on statistics, maybe more advanced topics
- applied guides: complete walk-throughs involving multiple elements

## Checklist

-   [ ] readthedocs build fails, fix
-   [ ] change approved
-   [ ] implementation finished
-   [ ] correct namespace imported
-   [ ] tests added
-   [ ] CHANGELOG updated


tgag17:
Thank you for following through with the issue! According to me, the newly added Tutorials section looks great. I had one doubt though. Should the `Installing zfit` and `5 minutes to zfit` sections in Getting Started have a link in the left column for easy navigation? The Getting Started section had the same earlier and I personally found them to be quite feasible.
Thanks!

mayou36:
Yes, good catch, I changed this and went over it a bit. Now it should be better overall. What do you think?

PR #371, opened 2021-11-13
KDE Boundary Mirror Logic 

schmitse:
Fixes #

## Proposed Changes

- Change the mirror logic of the `KDE1DimExact`, `KDE1DimGrid`, `KDE1DimFFT`, and `KDE1DimISJ` to use the boundary of the observable space instead of the minimum/maximum of the data. This prevents unwanted ''bumps'' in the pdf when the data falls to zero inside of the observable range and both edges are mirrored. 
- Updated the bandwidth description in the docstring of the `KDE1DimExact` and `KDE1DimGrid` classes to correctly reflect on the available bandwidth options. 

## Tests added

- Added tests to the `test_kde_funcs.py` 

## Checklist

- [x] change approved
- [x] implementation finished
- [x] tests added
- [x] CHANGELOG updated


zfit/models/kde.py
```diff
@@ -428,7 +428,7 @@ def _convert_init_data_weights_size(self, data, weights, padding):
         if callable(padding):
             data, weights = padding(data=data, weights=weights)
         elif padding is not False:
-            data, weights = padreflect_data_weights_1dim(data, weights=weights, mode=padding)
+            data, weights = padreflect_data_weights_1dim(data, weights=weights, mode=padding, limits=None)
```
jonas-eschle, line 431:
```suggestion
            data, weights = padreflect_data_weights_1dim(data, weights=weights, mode=padding, limits=limits)
```
this should be limits, right?

zfit/models/kde.py
```diff
@@ -428,7 +428,7 @@ def _convert_init_data_weights_size(self, data, weights, padding):
         if callable(padding):
             data, weights = padding(data=data, weights=weights)
```
jonas-eschle, line 429:
```suggestion
            data, weights = padding(data=data, weights=weights, limits=limits)
```
Let's give here limits too  to be consistent

zfit/models/kde.py
```diff
@@ -1363,7 +1371,7 @@ def __init__(self,
                 raise ValueError("obs can only be None if data is ZfitData with limits.")
             else:
                 obs = data.space
-        data, size, weights = self._convert_init_data_weights_size(data, weights, padding=padding)
+        data, size, weights = self._convert_init_data_weights_size(data, weights, padding=padding, limits=obs.limits)
```
jonas-eschle, line 1374:
```suggestion
        data, size, weights = self._convert_init_data_weights_size(data, weights, padding=padding, limits=obs)
```

This should just be "obs", it's already a space, right? Or do I mix up something?

jonas-eschle:
Hey looks good! I may have missed something in the logic though, but that can be on my side

zfit/models/kde.py
schmitse: I think its correct this way as padreflect wants to obtain the numerical limits not the space.

jonas-eschle: Indeed, you're right!

jonas-eschle:
That seems fine so far. If you think the implementation is finished, just remove the WIP in the name and request a review. Make sure to tick of the boxes, so you see that everything is done (e.g. add an entry in the CHANGELOG)

jonas-eschle:
Hey, was just wondering what's the status of this?

schmitse:
Hi im sorry for the stall. im on it today again. 
Fell ill and had work related deadlines coming up.

schmitse:
In the changelog i have added this: 
```
- KDE datasets are now mirrored around observable space limits
```
to the bugfixes/small changes section. I cannot `git add` it though, am i supposed to `git add -f` the changelog or how does it work?

schmitse:
Some coverage tests are showing up as failed. 
If i interpret the results correctly, for example from `coverage/coveralls` then the complaint it from the file: `zfit/minimizers/minimizers_scipy.py`, which i did not modify? 
How can i fix this/ Is there a way i can fix this, @jonas-eschle ? :)

jonas-eschle (approved):
Thanks a lot, this looks good! LGTM

PR #404, opened 2022-05-31
build: set zfit version with importlib-metadata

redeboer:
[`importlib.metadata`](https://docs.python.org/3/library/importlib.metadata.html) is preferred over the older `pkg_resources`. The current setup with `pkg_resources` [also causes CI problems](https://github.com/ComPWA/compwa-org/runs/6612064353?check_suite_focus=true#step:5:210).

See also https://github.com/zfit/phasespace/issues/80

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,7 @@ deprecated
 dotmap
 hist
 iminuit>=2.3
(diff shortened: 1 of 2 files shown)


jonas-eschle:
Thanks a lot! LGTM

PR #405, opened 2022-05-31
Made _pdf and _integrate methods cacheable

mashelll:
Fixes #

## Proposed Changes

- Was added class CacheablePDF that made pdf and integrate methods cacheable

## Tests added

-

## Checklist

- [ ] change approved
- [ ] implementation finished
- [ ] tests added
- [ ] CHANGELOG updated


tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,22 @@
...
+from zfit.models.cache import CacheablePDF
+
+mu = zfit.Parameter("mu", 1.0, -5, 5)
+sigma = zfit.Parameter("sigma", 1, 0, 10)
+obs = zfit.Space("x", limits=[-5.0, 5.0])
+gauss = zfit.pdf.Gauss(mu=mu, sigma=sigma, obs=obs)
+cached_gauss = CacheablePDF(gauss)
+x = znp.linspace(-5, 5, 500)
+
+
```
jonas-eschle, line 14: Also, extend the tests. We want to check if it actually caches and if it still works when the parameters are changed

zfit/models/cache.py
```diff
@@ -0,0 +1,111 @@
...
+        )
+        self.pdf_cache = None
+        self.cached_integral_limits = None
+        self.integral_cache = None
+        self.integral_cache_valid = tf.Variable(initial_value=False, trainable=False)
+
+        self.cache_tolerance = 1e-8 if cache_tol is None else cache_tol
+
+    @supports(norm="space")
+    def _pdf(self, x, norm_range):
```
jonas-eschle, line 45: use "norm" instead of `norm_range`

jonas-eschle (changes requested):
This looks good! What we need now is documentation (in the init of the Cacheable PDF) and extending the tests

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,22 @@
...
+import zfit.z.numpy as znp
+import tensorflow as tf
+from zfit.models.cache import CacheablePDF
+
+mu = zfit.Parameter("mu", 1.0, -5, 5)
+sigma = zfit.Parameter("sigma", 1, 0, 10)
+obs = zfit.Space("x", limits=[-5.0, 5.0])
+gauss = zfit.pdf.Gauss(mu=mu, sigma=sigma, obs=obs)
+cached_gauss = CacheablePDF(gauss)
+x = znp.linspace(-5, 5, 500)
```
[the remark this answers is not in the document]
mashelll: fixed

jonas-eschle:
@mashelll I've just updated the branch, so please pull first before you push again

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+    _PARAMS = ["alpha"]
+
+    def __init__(self, obs: ztyping.ObsTypeInput, **params):
+        super().__init__(obs, **params)
+        self.pdf_call_counter = tf.Variable(0.0)
+        self.integrate_call_counter = tf.Variable(0.0)
+        _mu = zfit.Parameter("mu", 1.0, -5, 5)
+        _sigma = zfit.Parameter("sigma", 1, 0, 10)
+        _obs = zfit.Space("x", limits=[-5.0, 5.0])
+        self._gauss = zfit.pdf.Gauss(mu=_mu, sigma=_sigma, obs=_obs)
```
jonas-eschle, line 20:
```suggestion
    def __init__(self, obs: ztyping.ObsTypeInput, mu, sigma):
        gauss = zfit.pdf.Gauss(mu=mu, sigma=sigma, obs=obs)
        super().__init__(obs, pdfs=gauss)
        self.pdf_call_counter = tf.Variable(0.0)
        self.integrate_call_counter = tf.Variable(0.0)

```

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+import zfit.z.numpy as znp
+import tensorflow as tf
+
+from zfit import supports
+from zfit.models.cache import CacheablePDF
+from zfit.util import ztyping
+
+
+class TestPDF(zfit.pdf.ZPDF):
+    _PARAMS = ["alpha"]
```
jonas-eschle, line 11:
```suggestion
class TestPDF(zfit.pdf.BaseFunctor):
```

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+        self.integrate_call_counter = tf.Variable(0.0)
+        _mu = zfit.Parameter("mu", 1.0, -5, 5)
+        _sigma = zfit.Parameter("sigma", 1, 0, 10)
+        _obs = zfit.Space("x", limits=[-5.0, 5.0])
+        self._gauss = zfit.pdf.Gauss(mu=_mu, sigma=_sigma, obs=_obs)
+
+    @supports(norm="space")
+    def _pdf(self, x, norm, *, norm_range=None):
+        self.pdf_call_counter.assign_add(1.0)
+        return self._gauss.pdf(x)
```
jonas-eschle, line 25:
```suggestion
        return self.pdfs[0].pdf(x)
```

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @supports(norm="space")
+    def _integrate(self, limits, norm, options=None):
+        self.integrate_call_counter.assign_add(1.0)
+        return self._gauss.integrate(limits)
```
jonas-eschle, line 30:
```suggestion
        return self.pdfs[0].integrate(limits)
```

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+
+def test_cached_pdf_equals_pdf_without_cache():
+    mu = zfit.Parameter("mu", 1.0, -5, 5)
+    sigma = zfit.Parameter("sigma", 1, 0, 10)
+    obs = zfit.Space("x", limits=[-5.0, 5.0])
+    gauss = zfit.pdf.Gauss(mu=mu, sigma=sigma, obs=obs)
+    cached_gauss = CacheablePDF(gauss)
+    x = znp.linspace(-5, 5, 500)
+
+    assert tf.math.reduce_all(tf.equal(gauss.pdf(x), cached_gauss.pdf(x)))
```
jonas-eschle, line 41:
```suggestion
    assert tf.math.reduce_all(tf.equal(gauss.pdf(x), cached_gauss.pdf(x)))
```
just to make sure the cache is used

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+
+
+def test_pdf_cache_is_used():
+    obs = zfit.Space("x", limits=[-5.0, 5.0])
+    alpha = zfit.Parameter("alpha", 1.0, -5, 5)
+    test_pdf = TestPDF(alpha=alpha, obs=obs)
```
jonas-eschle, line 47:
```suggestion
    mu = zfit.Parameter("mu", 1.0, -5, 5)
    sigma = zfit.Parameter("sigma", 1, 0, 10)
    test_pdf = TestPDF(obs=obs, mu=mu, sigma=sigma)
```

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+    cached_test_pdf.pdf(x)
+    assert tf.equal(test_pdf.pdf_call_counter, tf.Variable(1.0))
+    cached_test_pdf.pdf(x)
+    assert tf.equal(test_pdf.pdf_call_counter, tf.Variable(1.0))
+
+
+def test_pdf_cache_revaluation_if_parameter_was_changed():
+    obs = zfit.Space("x", limits=[-5.0, 5.0])
+    alpha = zfit.Parameter("alpha", 1.0, -5, 5)
+    test_pdf = TestPDF(alpha=alpha, obs=obs)
```
jonas-eschle, line 60: same here as above

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+from zfit.util import ztyping
+
+
+class TestPDF(zfit.pdf.ZPDF):
+    _PARAMS = ["alpha"]
+
+    def __init__(self, obs: ztyping.ObsTypeInput, **params):
+        super().__init__(obs, **params)
+        self.pdf_call_counter = tf.Variable(0.0)
+        self.integrate_call_counter = tf.Variable(0.0)
```
jonas-eschle, line 16: Yes, that's a good way to go! With this we can count the calls.

However, don't use floats! Use ints. By simply using `0` instead of `0.0` (and/or specify the dtype to be of type int).
The point is that later on we will compare it to counts and floating point comparisons are usually not reliable

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+    alpha.set_value(2.0)
+    cached_test_pdf.pdf(x)
+    assert tf.equal(test_pdf.pdf_call_counter, tf.Variable(2.0))
+
+
+def test_pdf_cache_revaluation_if_x_was_changed():
+    obs = zfit.Space("x", limits=[-5.0, 5.0])
+    alpha = zfit.Parameter("alpha", 1.0, -5, 5)
+    test_pdf = TestPDF(alpha=alpha, obs=obs)
+    cached_test_pdf = CacheablePDF(test_pdf)
```
jonas-eschle, line 75: same as above, maybe refactor this into a function that returns you the PDF?

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+
+def test_cached_integrate_equals_integrate_without_cache():
+    mu = zfit.Parameter("mu", 1.0, -5, 5)
+    sigma = zfit.Parameter("sigma", 1, 0, 10)
+    obs = zfit.Space("x", limits=[-5.0, 5.0])
+    gauss = zfit.pdf.Gauss(mu=mu, sigma=sigma, obs=obs)
+    cached_gauss = CacheablePDF(gauss)
+
+    assert tf.math.reduce_all(
+        tf.equal(gauss.integrate(limits=obs), cached_gauss.integrate(limits=obs))
```
jonas-eschle, line 93: Also, call it multiple times. Vary the limits and check if still works

tests/test_cacheable_pdf.py
```diff
@@ -0,0 +1,132 @@
...
+    obs = zfit.Space("x", limits=[-5.0, 5.0])
+    alpha = zfit.Parameter("alpha", 1.0, -5, 5)
+    test_pdf = TestPDF(alpha=alpha, obs=obs)
+    cached_test_pdf = CacheablePDF(test_pdf)
+    assert tf.equal(test_pdf.integrate_call_counter, tf.Variable(0.0))
+    cached_test_pdf.integrate(limits=obs)
+    assert tf.equal(test_pdf.integrate_call_counter, tf.Variable(1.0))
+    alpha.set_value(2.0)
+    cached_test_pdf.integrate(limits=obs)
+    assert tf.equal(test_pdf.integrate_call_counter, tf.Variable(2.0))
```
jonas-eschle, line 119: Good! Just make another call to it and check that it stays the same. The more checks, the better

jonas-eschle (changes requested):
Good! I am still thinking of a way to check for the data, I guess I have an idea, but good!

You can directly press "commit suggestion" if you agree, no need to retype things.

jonas-eschle:
And sure, let's see how far we go! worst case it will be in review close to being merged, but I think this will make it! Also, ping me here once you made changes (maybe say in a line or two what you did). I should usually see this fast

mashelll:
fixes are done

zfit/models/cache.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def use_cache():
+        return cache
+
+    val = tf.cond(flag, use_cache, autoset_func)
+    return val
+
+
+class CacheablePDF(BaseFunctor):
+    def __init__(self, pdf, cache_tolerance=None, **kwargs):
```
jonas-eschle, line 22: This needs documentation. On an example, see another PDF. It should describe the behavior, arguments of the PDF

jonas-eschle (changes requested):
Good, that looks fine so far!

I'll make some changes for the dataset in order to only cache if the dataset is the same too.

Add some documentation into the PDF, add it in the changelog.rst and add it in the documentation (do you know where? In the docs, rst file. Try to do the same as for other PDFs, like SumPDF)

jonas-eschle (changes requested):
I'm adding the last missing piece of the puzzle: the hashing of datasets.

Each data has now an attribute `hashint`, which is either an integer or `None`. The integer (which is a Python integer) needs to be remembered too!
If it's None (check with `x.hashint is None`), then don't cache.

Makes sense? The PR should be merged soon, you can already add the code changes

mashelll:
Documentation was updated

jonas-eschle:
Okay very good! Now we just need to check in the caching for the dataset, do you know how to do it? Or would you need assistance?

PR #407, opened 2022-06-25
Upd Monte-Carlo integration (vegasflow package)

mensigo:
Fixes #

## Proposed Changes
Add the 2nd MC method based on vegas algorithm (tensorflow implementation).
Extra requirements: vegasflow, joblib.
-

## Tests added
test_mc_vf_integration (3 simple functions, fully integrated)
-

## Checklist

- [ ] change approved
- [ ] implementation finished
- [ ] tests added
- [ ] CHANGELOG updated

### DIFF
--- a/tests/test_integrate.py
+++ b/tests/test_integrate.py
@@ -14,6 +14,9 @@ from zfit.core.parameter import Parameter
 from zfit.core.space import Space
 from zfit.models.dist_tfp import Gauss
 
+np.random.seed(42)
+tf.random.set_seed(42)
+
 limits1_5deps = [((1.0, -1.0, 2.0, 4.0, 3.0),), ((5.0, 4.0, 5.0, 8.0, 9.0),)]
 limits_simple_5deps = [((1.0, -1.0, -5.0, 3.4, 2.1),), ((5.0, 5.4, -1.1, 7.6, 3.5),)]
 
@@ -246,7 +249,7 @@ def func4_3deps_1_integrated(x, limits):
 @pytest.mark.parametrize("chunksize", [10000000, 1000])
 @pytest.mark.parametrize("limits", [limits2, limits2_split])
 def test_mc_integration(chunksize, limits):
-    # simpel example
+    # simple example
     zfit.run.chunking.active = True
     zfit.run.chunking.max_n_points = chunksize
     num_integral = zintegrate.mc_integrate(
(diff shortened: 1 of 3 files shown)


zfit/core/integration.py
```diff
@@ -80,6 +82,34 @@ def auto_integrate(
...
         integral = simpson_integrate(func=func, limits=limits, num_points=num_points)
+    elif method.lower() == "mc_vegasflow":
+        vf_options = vf_options or {}
+        n_iter = vf_options.get("n_iter", 20)
+        n_iter_pre = vf_options.get("n_iter_pre", 5)
+        n_events = vf_options.get("n_events", int(6e6))
+        events_limit = vf_options.get("events_limit", int(1e6))
+        list_devices = vf_options.get("list_devices")
+        compilable = vf_options.get("compilable", True)
+        verbose = vf_options.get("verbose", False)  # get_verbosity() >= 0 (?)
```
jonas-eschle, line 93: Yep, we can use the global zfit settings. What are the different verbosity levels in vegasflow? zfit goes from 0 to 10

tests/test_integrate.py
```diff
@@ -280,6 +312,60 @@ def test_mc_integration(chunksize, limits):
     ).numpy() == pytest.approx(integral3, rel=0.03)
 
 
+def test_mc_vf_integration():
+    # simple examples
+    num_integral = zintegrate.mc_vf_integrate(
```
jonas-eschle, line 317: There comes a crucial thing: can this run with a `z.function` wrapped? Because in the PDF, it usually will be.
You can also swap the default integration methods for PDFs and check if the tests still run, do they?

zfit/core/integration.py
```diff
@@ -339,6 +369,126 @@ def print_none_return():
...
+    Returns:
+        The integral
+    """
+    os.environ["VEGASFLOW_LOG_LEVEL"] = kwargs.get(
+        "VEGASFLOW_LOG_LEVEL", "2" if verbose else "1"
+    )
+    os.environ["VEGASFLOW_FLOAT"] = kwargs.get("VEGASFLOW_FLOAT", "64")
+    os.environ["VEGASFLOW_INT"] = kwargs.get("VEGASFLOW_INT", "32")
+    if list_devices is None:
+        list_devices = ["GPU"] if tf.config.list_physical_devices("GPU") else ["CPU"]
```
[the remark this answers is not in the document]
mensigo: + Actually, it's not needed. Removed it.

zfit/core/integration.py
```diff
@@ -339,6 +369,126 @@ def print_none_return():
...
+
+            def vf_integrand(x):
+                """Same vf_integrand, but n_dim > 1."""
+                new_x = lower + (upper - lower) * x
+                jac = tf.reduce_prod(upper - lower)
+                return func(new_x) * jac
+
+        if compilable:
+            spec_shape = (None,) if n_dim == 1 else (None, n_dim)
+            vf_integrand = tf.function(
```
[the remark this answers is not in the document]
mensigo: \+

zfit/core/integration.py
mensigo: I think, it's better to use vf's verbosity levels for convenience (4 levels) – added to docstring. However, it can be sync with zfit's levels in the future.

tests/test_integrate.py
mensigo: I tried wrapped version and faced some issues. The thing is, vf's `run_integration` is not intended to run under tf.function – there are some python control flow instructions (like "if" in `print_iteration` and numpy math ops when computing results from iterations) and tf.Variable creation (which causes "ValueError: tf.function-decorated function tried to create variables on non-first call"). I tried to use tf.py_function, but it did not help (anyway, i think it's not a relevant option in case of graph-like z.function application). I keep on thinking how to handle this issue.

zfit/core/integration.py
jonas-eschle: The problem is that it's an additional effort to get this right. But it can also be called from outside, yeah

tests/test_integrate.py
jonas-eschle: Hm, yep. that's a bit of a problem. One thing you can try is to run it in a [`tf.init_scope`](https://www.tensorflow.org/api_docs/python/tf/init_scope) or just make an issue with VF/ask them. That would be good to be solved somehow

tests/test_integrate.py
```diff
@@ -243,10 +247,38 @@ def func4_3deps_1_integrated(x, limits):
...
+limits5 = [(tuple(np.zeros(8)),), (tuple(np.ones(8)),)]
+func5_mean, func5_sigma = 0.5, 0.01
+func5_prod_gauss_lst = []
+for i in range(1, 9):
+    func5_obs_i = zfit.Space(f"gauss_{i}", limits=(0, 1))
+    func5_mu_i = zfit.Parameter(f"mu_{i}", func5_mean)
+    func5_sigma_i = zfit.Parameter(f"sigma_{i}", func5_sigma)
+    func5_gauss_i = zfit.pdf.Gauss(obs=func5_obs_i, mu=func5_mu_i, sigma=func5_sigma_i)
+    func5_prod_gauss_lst.append(func5_gauss_i)
+func5_prod_gauss = zfit.pdf.ProductPDF(pdfs=func5_prod_gauss_lst[::-1])
```
jonas-eschle, line 259: Move all of this inside a function

zfit/core/integration.py
```diff
@@ -378,8 +374,8 @@ def mc_vf_integrate(
     n_events: int = int(6e6),
     events_limit: int = int(1e6),
     list_devices: list = None,
-    compilable: bool = True,
-    verbose: bool = False,
+    compilable: bool = False,
+    verbose: int | str = 3,
```
jonas-eschle, line 378:
```suggestion
    verbose: int | str = 0,
```
Let the default be quiet

PR #445, opened 2023-02-24
Draft: Fix conv method

tinnuche:
I had a play and my initial work actually introduced some bugs, however I learnt a few things and I have figured out what I want to add:

- Some examples on the zfit docs
- Different obs: one shared for kernel / func ( different ranges ), another for the final output
- Currently the convolution method is "clever": it automatically considers leakage from outside the specificed func region. This is good, but I think to expand utitility, we should allow the option to to only convolve the specified regions, as this can be used more easily for multi-regions.
- allow use of kernel which is larger than func -> even if this requires the user to swap them around, so long as this does work


zfit/models/convolution.py
```diff
@@ -240,6 +243,10 @@ def __init__(
             "nbins_func": nbins_func,
         }
 
+        ### define subcomponent obs
+        self.obs_func = func.space
+        self.obs_kernel = kernel.space
```
jonas-eschle, line 248:
```suggestion
        self._space_func = func.space
        self._space_kernel = kernel.space
```

zfit/models/convolution.py
```diff
@@ -255,12 +262,12 @@ def _unnormalized_pdf(self, x):
         x_func = tf.meshgrid(*tf.unstack(x_funcs, axis=-1), indexing="ij")
         x_func = znp.transpose(x_func)
         x_func_flatish = znp.reshape(x_func, (-1, self.n_obs))
-        data_func = Data.from_tensor(tensor=x_func_flatish, obs=self.obs)
+        data_func = Data.from_tensor(tensor=x_func_flatish, obs=self.obs_func)
```
jonas-eschle, line 265:
```suggestion
        data_func = Data.from_tensor(tensor=x_func_flatish, obs=self._space_func)
```

zfit/models/convolution.py
```diff
@@ -255,12 +262,12 @@ def _unnormalized_pdf(self, x):
...
 
         x_kernel = tf.meshgrid(*tf.unstack(x_kernels, axis=-1), indexing="ij")
         x_kernel = znp.transpose(x_kernel)
         x_kernel_flatish = znp.reshape(x_kernel, (-1, self.n_obs))
-        data_kernel = Data.from_tensor(tensor=x_kernel_flatish, obs=self.obs)
+        data_kernel = Data.from_tensor(tensor=x_kernel_flatish, obs=self.obs_kernel)
```
jonas-eschle, line 270:
```suggestion
        data_kernel = Data.from_tensor(tensor=x_kernel_flatish, obs=self._space_kernel)
```

jonas-eschle (commented):
Thanks, that's a good fix! I renamed them appropriately and made the attributes private -> meaning it's not a "stable API", i.e. users should not rely on this attribute to be there, it's an implementation detail.

jonas-eschle:
> found serious issue that if space of kernel and space of func are not the same, then the convolution doesnt work properly:

Hm, not sure, the example doesn't run for me but fails with an error that the obs don't match: this is expected as the convolution is one-dimensional and therefore needs PDFs that are defined in the same variables.

> Want zfit conv method to work with func and kernel and conv pdfs having 3 seperate obs.

So let's say, it would work on each part separately? I think you'll need to adjust the conv ranges larger, overlapping. If you then "cut" (mentally, currently) the PDFs at the right spot and glue them together, does that work?

PR #510, opened 2024-02-28
feat: add different sigmas on each side of the double crystal ball pdf

iasonkrom:
PR to make each side of the double crystal ball be able to have a different standard deviation for the gaussian

- [x] change approved
- [x] implementation finished
- [x] tests added
- [x] CHANGELOG updated


iasonkrom:
The tests for the crystal ball pass locally for now when setting `sigmal=sigmar`. This is still WIP and we gotta think if and how we wanna have a version where the left and the ride share one `sigma` in the `__init__` method.

iasonkrom:
I actually suggest we use different underlying functions for `DoubleCB` and `GeneralizedCB` until https://github.com/zfit/zfit/issues/511 is fixed because if we don't do that, `DoubleCB` is broken with tensorflow 2.15. The extra functions can be removed when that issue is fixed upstream. I already did the change

iasonkrom:
I added the new `GeneralizedCB` to `tests/serialize/test_basics.py` but the `.asdf` files are missing. How to I make them?

zfit/models/physics.py
```diff
@@ -393,3 +462,129 @@ class DoubleCBPDFRepr(BasePDFRepr):
...
+        sigmar: ztyping.ParamTypeInput,
+        alphar: ztyping.ParamTypeInput,
+        nr: ztyping.ParamTypeInput,
+        obs: ztyping.ObsTypeInput,
+        *,
+        extended: ExtendedInputType = None,
+        norm: NormInputType = None,
+        name: str = "GeneralizedCB",
+    ):
+        """Generalized Double-sided Crystal Ball shaped PDF. A combination of two CB using the **mu** (not a frac) and a
```
jonas-eschle, line 485:
```suggestion
        """Generalized asymetric double-sided Crystal Ball shaped PDF. A combination of two CB using the **mu** (not a frac) and a
```

jonas-eschle (approved):
Looks good to me, you've seem to have perfectly well understood what to change where! Don't forget to add yourself to the contributors in the AUTORS.rst, but other than that, it's fine from my side. Let me know when you want it merged

iasonkrom:
> I added the new `GeneralizedCB` to `tests/serialize/test_basics.py` but the `.asdf` files are missing. How to I make them?

I want to fix this thing before merging

jonas-eschle:
Ah yes, you can run the tests with `--recreate-truth`, which will create new "truth" files

iasonkrom:
Gothca. Test passes locally. Do I just commit the recreated and new truths?

jonas-eschle:
Exactly, they're also visible in the PR as the diff (which also makes sure the rest of the truth isn't changed)

iasonkrom:
hmm, I don't know if these truth changes that occured are fine. Please check those. Other than that, in terms of source code, I'm happy with merging.

iasonkrom:
The non crystal ball distribution tests were passing before these truth changes on my machine

jonas-eschle:
Yes, I've just checked them, I highly suspect it's because of the asdf upgrade from 2 to 3, it all looks sane to me

jonas-eschle:
Just checking for the tests (some will fail, aware of that, some timeouts for certain binned stuff in eager mode), but looks good, then I'll merge!

iasonkrom:
@jonas-eschle is the test that is still running broken or does it really take very very long

jonas-eschle:
Alright, the ones still running are slow but fine, that's a different story, the crucial ones passed. LGTM!

PR #517, opened 2024-03-09
enh: allow parameters with same name to co-exist

jonas-eschle:
## Proposed Changes

- Parameters can now have the same name

## Tests added

- checking for new behavior that same names are allowed if clearly defined.

## Checklist

- [x] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated


ikrommyd:
Hi @jonas-eschle, should this fix the issue we had in the `GeneralizedCB` where using the `GeneralizedCB` with the same `sigma` for the left and right side to describe the `DoubleCB` was failing?

jonas-eschle:
Hey @ikrommyd , no, this is actually a significantly longer standing "issue" and has to do with the definition of parameters and uniqueness, in short, zfit didn't allow to create more than one parameter with the same name. An extensive discussion is here: https://github.com/zfit/zfit/discussions/342

The problem with the same params is still open, I am currently cleaning up the repo and collecting remaining issues, trying to get a reproducible example of what we've seen

PR #518, opened 2024-03-11
feat: implement Voigt profile

ikrommyd:
Starting a PR to implement the Voigt Profile. The Faddeeva function is also gonna be implemented in order to do that.
For now it's just an empty commit but starting it early to get public feedback whilst doing it.

## Proposed Changes

- Implement the Faddeeva function
- Implement Voigt profile distribution

## Tests added

- Modify `tests/serialize/test_basics.py` and `tests/test_z_numpy.py` accordingly and add `tests/test_pdf_voigt.py`

## Checklist

- [x] change approved
- [x] implementation finished
- [x] tests added
- [x] CHANGELOG updated


ikrommyd:
@jonas-eschle I added the Faddeeva function implementation into the znp module by making a few changes. Please review those before moving to the Voigt implementation. Of course tests will be added but I'm just looking to see if you like the structure of the znp module that way.

jonas-eschle:
Looks good! We could also convert it to a module, but it really doesn't matter, perfectly fine this way
For the name, why not name it directly faddeeva? And maybe we could call it faddeevafast or similar? Just to stay forward compatible if we want to have a high precisiosn variante. But it could aso be an argument and we dispatch if needed.
Anyways, I would have just one name, not too if there is no reason to have alisases

One thing, the function should not be tf compiled by default: if a user would use pure python objects, it would trigger a recompilation every time (the tf builtings aren't compiled either).

ikrommyd:
I will probably just name it `faddeevafast` or something like that only. Just wanted to keep the name originally as a homage to the original authors but we can just do it in the docstring.

For the compilation, should I just remove the decorator completely and the user can just decorate the function however they want in their one code? If so, do we want to decorate it in some manner when being called through the Voigt profile though?

jonas-eschle:
> I will probably just name it `faddeevafast` or something like that only. Just wanted to keep the name originally as a homage to the original authors but we can just do it in the docstring.

Yes, the docstring, or at least the source code, should ideally contain that. Docstring should most importantly cite the method in one way or the other

> For the compilation, should I just remove the decorator completely and the user can just decorate the function however they want in their one code? If so, do we want to decorate it in some manner when being called through the Voigt profile though?

Yes, that makes sense, if we can control to give in just `znp.array` objects. That means it won't just recompile. This can happen right within the file where the PDF itself is defined (and on the voight probably could be enough). The difference should be marginal, however, as usually the outer function is compiled anyways

ikrommyd:
This is how we're currently doing.
image

zfit/z/numpy.py
```diff
@@ -10,14 +10,140 @@
...
-    inv = staticmethod(tf.linalg.inv)
-    det = staticmethod(tf.linalg.det)
-    solve = staticmethod(tf.linalg.solve)
+    inv = staticmethod(_tf.linalg.inv)
+    det = staticmethod(_tf.linalg.det)
+    solve = staticmethod(_tf.linalg.solve)
 
 
-del tf
+def faddeeva_humlicek(z, s=10.0):
```
jonas-eschle, line 24: few things: it's probably good to move it under a "special" namespace, but we don't have that right now. For future, maybe make a comment?
document `s` (and maybe also replace it in the formula below?)

zfit/z/numpy.py
```diff
@@ -10,14 +10,140 @@
...
+        (
+            (((((ZZ + bb[7]) * ZZ + bb[6]) * ZZ + bb[5]) * ZZ + bb[4]) * ZZ + bb[3])
+            * ZZ
+            + bb[2]
+        )
+        * ZZ
+        + bb[1]
+    ) * ZZ + bb[0]
+
+    w = _tf.tensor_scatter_nd_update(w, _tf.where(mask), numer / denom)
```
jonas-eschle, line 147:
```suggestion
    w = _tf.tensor_scatter_nd_update(w, where(mask), numer / denom)
```
where is in znp namespace

jonas-eschle (changes requested):

jonas-eschle:
Great performance comparison! Ideally, you could add this in the test/a benchmark folder in the tests and save the plot. If it's expensive, it can also just be rerun if the --plot argument is given to the tests

zfit/z/numpy.py
ikrommyd: I actually reverted this commit. Tensorflow `where` is different from the numpy `where`. We want the tensorflow one there.

zfit/z/numpy.py
ikrommyd: Fixed. Is there a good place to also add the Faddeeva function in the docs?

ikrommyd:
I had to do `--recreate-truth` for `tests/serialize/test_basics.py` so please check the new truths. The tests that weren't involving the Voigt were passing before recreating the truths.

ikrommyd:
Some more testing for the Voigt should be added and maybe some explicit testing of the faddeeva function. We can actually do comparisons vs scipy for both of those. But anyways I removed the draft from the PR as it should be more or less ready.

ikrommyd:
@jonas-eschle I'm happy with the testing of Voigt against scipy that I just added. Let me know if you want something more in that domain.
I think it would be a good idea to jit compile the Faddeeva function within the Voigt PDF and always makes sure that what gets passed into the Faddeeva function is `tf.Tensor` which is already there since we use `znp.asarray` on the input

tests/test_z_numpy.py
```diff
@@ -14,3 +17,57 @@ def test_z_numpy_ndarray_is_tensorflow_tensor():
...
+        rtol=1e-06,
+        atol=1e-10,
+    )
+
+    test_values_tensor = znp.asarray(test_values)
+    with tf.GradientTape(persistent=True) as tape:
+        tape.watch(test_values_tensor)
+        result = znp.faddeeva_humlicek(test_values_tensor)
+        gradientstf = tape.gradient(result, test_values_tensor)
+    assert np.all(np.isfinite(gradientstf))
```
ikrommyd, line 73: @jonas-eschle what would be the proper assertion here to check the gradient.

jonas-eschle (changes requested):

jonas-eschle:
> Some more testing for the Voigt should be added and maybe some explicit testing of the faddeeva function. We can actually do comparisons vs scipy for both of those. But anyways I removed the draft from the PR as it should be more or less ready.

Yes, a test for the faddeeva vs wovf would make sense, so we're testing things independently

> I think it would be a good idea to jit compile the Faddeeva function within the Voigt PDF and always makes sure that what gets passed into the Faddeeva function is tf.Tensor which is already there since we use znp.asarray on the input

This is already taken care of by the PDF itself: it preprocesses the data, converts it to `znp.array`, and compiles, if needed

jonas-eschle:
For the name of the Voight(/Cauchy) mean, I didn't really come up with something else, everything seems as good or as bad, so just `m` is probably fine, it can also stand for `mean` and is consistent. So fine with the `m`

ikrommyd:
> This is already taken care of by the PDF itself: it preprocesses the data, converts it to `znp.array`, and compiles, if needed
Ah so no need to wrap the faddeeva explicitly around a `tf.function(..., jit_compile=True)` cool.

tests/test_z_numpy.py
ikrommyd: I don't really understand the equivalency here. The `z.math` functions take in a func with no arguments and zfit params. `zfit.Parameter` converts my complex input tensor into a float one and I want to check the gradients of `w(z)` with respect to `z` for a complex `z` tensor.

ikrommyd: We can test the derivatives of the `Voigt` with respect to `m`, `sigma` and `gamma` this way though

jonas-eschle:
> Ah so no need to wrap the faddeeva explicitly around a `tf.function(..., jit_compile=True)` cool.

yes, zfit has a bit more fine-grained control. A mixture of "better" and a bit of "legacy", but overall, it compiles it if it's expensive (i.e. if you run a fit)

tests/test_z_numpy.py
jonas-eschle: > We can test the derivatives of the `Voigt` with respect to `m`, `sigma` and `gamma` this way though

yes, sorry, that's what I had in mind, I should have written it too. For x, one could use a numerical comparison, (like the numdifftools in `z.math`) or also just check for it to be finite, that should be the main possible issue

zfit/z/numpy.py
jonas-eschle: Ah yes? It's just the shape though I think?

zfit/z/numpy.py
jonas-eschle: Hm, there is currently no default case, but that's a bit historic as the "new backend", znp, is introduced as a "works if works", as it's still experimental in TF, technically. But it was so stable, factually, that we exposed it.
That would need to add a backend API docs file, if you want to try, please go ahead and add what you think is sensibse (that would mean to add the zfit.z.numpy in the `user_api`, otherwise, I'll note it down for the docs overhaul

tests/test_z_numpy.py
```diff
@@ -14,3 +17,57 @@ def test_z_numpy_ndarray_is_tensorflow_tensor():
...
+            test_values2,
+            test_values3,
+            test_values4,
+            test_values5,
+            test_values6,
+            test_values7,
+            test_values8,
+        ]
+    )
+    assert np.allclose(
```
jonas-eschle, line 61:
```suggestion
    np.testing.assert_allclose(
```

zfit/models/basic.py
```diff
@@ -240,3 +241,98 @@ class ExponentialPDFRepr(BasePDFRepr):
...
+        name: str = "Voigt",
+    ):
+        """Voigt profile.
+
+        The Voigt profile is a convolution of a 1-D Gaussian distribution with standard deviation sigma and a 1-D Cauchy distribution with half-width at half-maximum gamma.
+
+        The (unnormalized) Voigt profile shape is defined as
+
+        .. math::
+            V(x; \\sigma, \\gamma) = Re[w(z)]
```
jonas-eschle, line 267:
```suggestion
        .. math::
        
            V(x; \\sigma, \\gamma) = Re[w(z)]
```
needs a space afaik (rest docs: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-math)

tests/test_pdf_voigt.py
```diff
@@ -0,0 +1,30 @@
...
+
+    mean = zfit.Parameter("mean", mean1_true, -10, 10)
+    sigma = zfit.Parameter("sigma", std1_true, 0.1, 10)
+    gamma = zfit.Parameter("gamma", width1_true, 0.1, 10)
+    voigt1 = zfit.pdf.Voigt(m=mean, sigma=sigma, gamma=gamma, obs=obs1)
+
+    probs1 = voigt1.pdf(x=test_values, norm=False).numpy()
+    assert all(probs1) > 0
+    sample1 = voigt1.sample(100).numpy()
+    assert len(sample1) == 100
```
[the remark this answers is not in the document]
ikrommyd: sample1 is an instance of `Data` and has no attribute `shape`. I will use `assert len(sample1.value()) == 100`

tests/test_z_numpy.py
ikrommyd: I will maybe add something more here and let you know once it's ready to merge.

ikrommyd:
@jonas-eschle some comments you left were on an older version of the code. I had already changed the testing scripts to use `np.testing` in most parts. I believe I have resolved everything except from one comment that I may want to add something.

ikrommyd:
@jonas-eschle added a bit more testing. I think I'm done here. Give it one more once-over and merge it whenever you're happy with the tests.

jonas-eschle (approved):
Looks good to me, LGTM!

(just FYI: the gradienttape has an option `watch_accessed_variables`, by default, it watches all variables. Also, if you take the gradient _inside_ the tape, it will record the gradient (i.e. second order) for that operation -> I've changed both things)

Issue #512, opened 2024-03-01
Addition of new PDFs


jonas-eschle:
Fill up the ranks of PDFs with the following:
- [ ] Cruiff PDF (zfit physics?)
- [ ] Berstein
- [ ] Novosibirsk (zfit physics?)
- [ ] JohnsonSU (implementation from Matthieu?)
- [ ] CBExGauss
- [ ] CMSShape (zfit-physics)

iasonkrom:
- [ ] Voigtian
- [ ] Tsallis-Hagedorn, a model for the minimum bias pT distribution
- [ ] Q-Gaussian

PasqualeAndreola:
As stated by @iasonkrom , I would be happy to see the Voigtian, too

anjabeck:
Hi! I'd like to make a start on this. Just checking that noone is working on it already?

iasonkrom:
Hi @anjabeck, I'm doing the Voigt, and I'm also gonna do the CMSShape and CBExGauss because I need them. You can start doing any other distribution if you want. I believe @jonas-eschle is happy to see PRs

jonas-eschle:
Hi @anjabeck , you're very welcome to pick a PDF and start making a PR, I'll update the list

ikrommyd:
@jonas-eschle I believe q-gaussian is different from the generalized normal? We should also add Student's t distribution which we can do from tensorflow-probability and then the q-gaussian is computed via a change of variables from the t-distribution.

ikrommyd:
@anjabeck would you be willing to do some more distributions for tf-probability?

jonas-eschle:
> @jonas-eschle I believe q-gaussian is different from the generalized normal? We should also add Student's t distribution which we can do from tensorflow-probability and then the q-gaussian is computed via a change of variables from the t-distribution. I updated the issue.

Agree, it's the student T that we can have to go fro the q-gaussian

anjabeck:
> @anjabeck would you be willing to do some more distributions for tf-probability?

I guess I'd rather try something new to learn stuff :)

ikrommyd:
> > @anjabeck would you be willing to do some more distributions for tf-probability?
> 
> I guess I'd rather try something new to learn stuff :)

Do you want to take a shot at the Bernstein polynomials?

ikrommyd:
[...]

PR #520, opened 2024-03-11
Add ChiSquared pdf

anjabeck:
## Proposed Changes
- add chi2 distribution from [tfp](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/Chi2)

## Tests added
- check chi2 mean and std deviation

## Checklist

- [x] change approved
- [x] implementation finished
- [x] tests added
- [x] CHANGELOG updated

related to #512


tests/test_pdf_chisquared.py
```diff
@@ -0,0 +1,26 @@
...
+import numpy as np
+import pytest
+
+import zfit
+from zfit import Parameter
+from zfit.models.dist_tfp import ChiSquared
+
+N_true = 2
+
+obs = zfit.Space(obs="Nobs", limits=(0, 200))
```
jonas-eschle, line 11: create the obs and test values inside the test function (and use znp.name instead of np.name wherever possible)

tests/test_pdf_chisquared.py
```diff
@@ -0,0 +1,26 @@
+#  Copyright (c) 2024 zfit
+import numpy as np
+import pytest
+
+import zfit
+from zfit import Parameter
+from zfit.models.dist_tfp import ChiSquared
```
jonas-eschle, line 7: Use `zfit.pdf` to access the pdf. And remove the imports, inside the test, you can directly use `zfit.pdf.name` or `zfit.Parameter`

zfit/models/dist_tfp.py
```diff
@@ -493,3 +493,60 @@ class PoissonPDFRepr(BasePDFRepr):
...
+        *,
+        extended: ExtendedInputType = None,
+        norm: NormInputType = None,
+        name: str = "ChiSquared",
+    ):
+        """ChiSquared distribution for a ndof degrees of freedom.
+
+        The chisquared shape for $d$ degrees of freedom is defined as
+
+        .. math::
```
jonas-eschle, line 514: An empty line is needed between the directive and the content (yes, RST is quite sensitive to many things: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-math)

zfit/models/dist_tfp.py
```diff
@@ -493,3 +493,60 @@ class PoissonPDFRepr(BasePDFRepr):
...
+        super().__init__(
+            distribution=distribution,
+            dist_params=dist_params,
+            obs=obs,
+            params=params,
+            name=name,
+            extended=extended,
+            norm=norm,
+        )
+
```
jonas-eschle, line 552: Can you add here a Repr (as for the others)? This makes the PDF serializable

tests/test_pdf_chisquared.py
```diff
@@ -0,0 +1,26 @@
...
+
+test_values = np.random.uniform(low=0, high=100, size=100)
+
+
+def test_chisquared():
+    N = Parameter("N", N_true)
+    chi2 = ChiSquared(obs=obs, ndof=N)
+
+    probs1 = chi2.pdf(x=test_values)
+    probs1 = probs1.numpy()
```
jonas-eschle, line 21:
```suggestion
```
Not needed, has no effect

tests/test_pdf_chisquared.py
```diff
@@ -0,0 +1,26 @@
...
+
+    samples = chi2.sample(10000).numpy()
```
jonas-eschle, line 23:
```suggestion
    samples = chi2.sample(10000)
```
`.numpy()` is not needed in general

jonas-eschle (changes requested):
Thanks for the addition! A few minor things, but otherwise looks good to me.

Two more things:
- add the PDF to the docs, under docs/user_api/pdf/basic.rst
- add yourself to the contributors in AUTHORS.rst

ikrommyd:
Please also test in `tests/serialize/test_basics.py`

tests/test_pdf_chisquared.py
anjabeck: Ok sure.
For my curiosity: they are global because I started from a copy of [test_pdf_poisson.pdf](https://github.com/zfit/zfit/blob/develop/tests/test_pdf_poisson.py). Is this intended there because there is more than one function accessing the values? Or is it just legacy and no-one's gotten around to change it?

anjabeck:
Thanks for the comments, I'll change them all.

But for my learning: almost all of the comments refer to stuff that I did because I started from a copy of [test_pdf_poisson.py](https://github.com/zfit/zfit/blob/develop/tests/test_pdf_poisson.py).
Was there a change of conventions (using `znp`, no need for `.numpy()` anymore) and no-one's gotten around to change it? Or is there something peculiar about the poisson vs chi2 that really requires different treatment?

jonas-eschle:
> Thanks for the comments, I'll change them all.

Great, thanks for the contribution!

> But for my learning: almost all of the comments refer to stuff that I did because I started from a copy of [test_pdf_poisson.py](https://github.com/zfit/zfit/blob/develop/tests/test_pdf_poisson.py?rgh-link-date=2024-03-13T10%3A40%3A50Z). Was there a change of conventions (using `znp`, no need for `.numpy()` anymore) and no-one's gotten around to change it? Or is there something peculiar about the poisson vs chi2 that really requires different treatment?

No, you're perfectly right that the Poisson does it "wrong", or less elegant. Over time, things got added and not all tests got the same care of updates

jonas-eschle:
> Please also test in `tests/serialize/test_basics.py`

Good point! You will need to rerun the tests with `--recreate-truth` (once every other test, except the newly created, passes). This will create a new truth file that you can add (that's not yet documented in the contributor guide, updating)

anjabeck:
About the serialization: when adding the chi2 pdf to the set of [basic_pdfs](https://github.com/anjabeck/zfit/blob/chisquared-pdf/tests/serialize/test_basics.py#L318) via something like this
```python
def chisquared(extended=None, **kwargs):
    import zfit

    N = zfit.Parameter("N_chi2", 2)
    obs = zfit.Space("obs", limits=(0, 3))
    return zfit.pdf.ChiSquared(obs=obs,ndof=N,extended=extended)
#     return zfit.pdf.ChiSquared(obs=obs, ndof=N, extended=extended, norm=zfit.Space("norm", limits=(-3, 3)))
```
does not work because there is a mismatch with the other pdfs which are defined in (-3,3).
I tried to add it through defining a new non-zero pdf class:
```python
class chisquared_allownegativex(zfit.pdf.ZPDF):
    _PARAMS = ['ndof']

    def _unnormalized_pdf(self, x):
        x = x.unstack_x(x)
        ndof = self.params['ndof']
        chi2 = zfit.pdf.ChiSquared(obs=self.space, ndof=ndof, extended=self.is_extended)
        return zfit.z.where(x < 0, 0, chi2.pdf(x))


def chisquared(extended=None, **kwargs):
    import zfit

    N = zfit.Parameter("N_chi2", 2)
    obs = zfit.Space("obs", limits=(-3, 3))
    return chisquared_allownegativex(obs=obs, ndof=N, extended=extended)
```
but (probably expectedly) this complains that it doesn't know how to serialize this object
```bash
E           zfit.serialization.serializer.SerializationTypeError: All distributions must be ZfitSerializable
```

I guess one could just change all the range to only allow positive values but maybe that's not ideal.
How are you testing the Poisson right now? I fail to find it here.

jonas-eschle:
The "obs" that has to be the same is an inherent shortcomming of the current serialization, you're absolutely right!

If you would ever want to build a class that uses other PDFs, you should have a look at the Functors, they wrap other pdfs and provide convenient helpers.
This is indeed not quite easy: What should we do if it's < 0, i.e. fail or just set to 0? (I'll check for the poisson tomorrorow what we've done there). What would you expect if you fit something?

For the serialization, I've noted that, so we could also merge without that until the issue is fixed (as serialization is a best-effort basis)

jonas-eschle:
Update, once #523 is merged, this should be fixed (just needs CI to pass)

anjabeck:
> This is indeed not quite easy: What should we do if it's < 0, i.e. fail or just set to 0? (I'll check for the poisson tomorrorow what we've done there). What would you expect if you fit something?

The chi2 and poisson distributions are very well defined mathematical objects with neat properties. These objects are undefined below 0 and I personally think the code should reflect this behaviour and lead to an error.
If a user _really_ needs a distribution that has the shape of a chi2/Poisson but moved to a negative regime, they should build their own pdf. This will not be a chi2/Poisson in the classical sense and may or may not have all the same nice properties but that's the user's problem.

For the serialization, I would expect that is saves the fitting range as well as the normalization range for every individual pdf (both are relevant to rebuild the pdf, no?). I've definitely had real-world (or rather HEP 😅) usecases of fits to a variable in different data sets which required different fit ranges.

jonas-eschle:
> If a user really needs a distribution that has the shape of a chi2/Poisson but moved to a negative regime, they should build their own pdf. This will not be a chi2/Poisson in the classical sense and may or may not have all the same nice properties but that's the user's problem.

Fair, or use a truncated one basically, something like that should work.

> For the serialization, I would expect that is saves the fitting range as well as the normalization range for every individual pdf 

yes, this was simply a bug and fixed now, I've just updated the branch, so make sure to pull first before you push again

anjabeck:
Thanks for merging the bugfix but I'm not sure it solves my original problem: how should I include the chi2 into the serialization tests? I see two options:

1. Test it as part of the basic_pdfs. This would require some gymnastics to create a custom function with normalization range from -3 to 3 (otherwise the sum/product functions complain).
2. Write an entirely new test for pdfs with restricted ranges. This could also include the poisson for which I think there is no serialization test yet.

I think (1) does not make a lot of sense because it wouldn't test the `Repr` object of the chi2 function but rather whatever is used in the custom one. What do you think?

jonas-eschle:
Ah, I see, fair! Actually, you can just make it dynamic, depending on the _original_ pdf space, i.e.
`lower, upper = pdf.space.limit1d` gives you the limits, then sample with those

anjabeck:
> Ah, I see, fair! Actually, you can just make it dynamic, depending on the _original_ pdf space, i.e. `lower, upper = pdf.space.limit1d` gives you the limits, then sample with those

I don't understand how this helps. The current test adds pdfs which are defined on (-3,3). I cannot add `zfit.pdf.ChiSquared` to this test because it cannot have a domain or normalisation range below 0.

jonas-eschle:
> The current test adds pdfs which are defined on (-3,3)

I meant you can make this dynamic that it does _no longer_ use -3, 3, but instead the space of the pdf, which can be different for each pdf. This could potentially have some issues when pdfs autobuild from other pdfs and the obs don't agree, in this case, it's also fine to make it (0, 3) for all, there's nothing inherently peculiar about the negative values.

Otherwise, let me know if this causes too much troubles, I can make the changes

anjabeck:
I reran all tests with `--recreate-truth` and a lot of the binaries changed. Not sure if this is expected?

ikrommyd:
@anjabeck Please run the test without recreating the truth first on your machine. If only the chi-squared related tests fail it's fine.

anjabeck:
> @anjabeck Please run the test without recreating the truth first on your machine. If only the chi-squared related tests fail, it's fine.

> Good point! You will need to rerun the tests with --recreate-truth (once every other test, except the newly created, passes). This will create a new truth file that you can add (that's not yet documented in the contributor guide, updating)

Can you make up your minds about the policy? ;)

ikrommyd:
I think we said the same thing. "once every other test, except the newly created, passes".

anjabeck:
But the other tests cannot pass given the binaries have not been updated to the new range of (-2.1,3) which is nothing I changed but comes from #524

ikrommyd:
Ah I didn't notice the newly created range. I'm sorry then. Binaries generally change whenever I recreate truths. I believe the diff is fine and @jonas-eschle will confirm.

anjabeck:
Ah no worries. I thought so too about the binaries but given that the `json` files had changed as well, I thought it's probably a valid diff?

jonas-eschle:
Thanks for the discussions both of you, the diffs look good, this was indeed my oversight in the other PR, some git stuff slipped through.
Looks good overall, only thing missing seems to be to update the CHANGELOG

jonas-eschle (approved):
Looks all good!

PR #530, opened 2024-03-20
feat: add double-sided Gaussian with exponential tails PDF

ikrommyd:
Early PR to add double-sided Gaussian with exponential tails PDF and `CBExGauss` factory

## Proposed Changes

- add double-sided Gaussian with exponential tails PDF and factory to construct `CBExGauss` PDF from this and the `GeneralizedCB`. Implementation from: https://github.com/cms-egamma/egm_tnp_analysis/blob/master/libCpp/RooCBExGaussShape.cc

## Tests added

- Modify tests/serialize/test_basics.py and tests/test_z_numpy.py accordingly and add tests/test_pdf_gaussexptail.py

## Checklist

- [x] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated


ikrommyd:
@jonas-eschle the analytic integral isn't hard in terms of math. We know the analytic integrals of Gaussians, we have the analytic integral of the Crystal Ball function, and the exponential tail is something of the form `znp.exp(-0.5 * abs_alpha**2) * znp.exp(n * (t + abs_alpha))` which is just an integral of the form `exp(A * t)` and the anti-derivative is easy. All this just needs to be put well into code and treat the piecewise functions properly.

zfit/models/physics.py
```diff
@@ -576,6 +576,101 @@ def generalized_gaussexptail_func(x, mu, sigmal, alphal, nl, sigmar, alphar, nr)
...
+        abs_sigma / n * znp.exp(-0.5 * znp.square(abs_alpha) + n * abs_alpha) * (znp.exp(n * tmax) - znp.exp(n * tmin))
+    )
+    gauss_integral_alpha_tmax = (
+        abs_sigma * sqrt_pi_over_two * (tf.math.erf(tmax / sqrt2) - tf.math.erf(-abs_alpha / sqrt2))
+    )
+    exp_integral_alpha_tmin = (
+        abs_sigma
+        / n
+        * znp.exp(-0.5 * znp.square(abs_alpha) + n * abs_alpha)
+        * (znp.exp(-n * abs_alpha) - znp.exp(n * tmin))
```
jonas-eschle, line 619: combine the exp terms?

zfit/models/physics.py
```diff
@@ -576,6 +576,101 @@ def generalized_gaussexptail_func(x, mu, sigmal, alphal, nl, sigmar, alphar, nr)
...
+    integral_left = gaussexptail_integral_func(
+        mu=mu, sigma=sigmal, alpha=alphal, n=nl, lower=lower, upper=upper_of_lowerint
+    )
+    left = tf.where(tf.less(mu, lower), znp.zeros_like(integral_left), integral_left)
+
+    lower_of_upperint = znp.maximum(mu, lower)
+    integral_right = gaussexptail_integral_func(
+        mu=mu, sigma=sigmar, alpha=-alphar, n=nr, lower=lower_of_upperint, upper=upper
+    )
+    right = tf.where(tf.greater(mu, upper), znp.zeros_like(integral_right), integral_right)
```
jonas-eschle, line 669: why exactly is this split needed and we can't do `tf.where(bool, left, right)`? Is it maybe with NaNs and grad?

ikrommyd: I just did the same thing that was done for the crystal ball case. Will check.

zfit/models/physics.py
ikrommyd: Yes, I will merge/combine stuff all over the place as soon as I'm sure it's correct.

zfit/models/physics.py
ikrommyd: This just says that if `mu` is greater than the upper limit, set the right integral to zero because it's outside our integration range. The equaivalent exists for the left integral as well and then you sum the two. Can you show me what you think can be removed here?

ikrommyd:
@jonas-eschle I believe you can have a look at this point. Most of the core parts are down and tested. Adding a `CBExGauss` factory is something we can do now.

ikrommyd:
Added a `CBExGauss` constructor. Let me know what you think.

zfit/models/physics.py
```diff
@@ -548,3 +548,400 @@ class GeneralizedCBPDFRepr(BasePDFRepr):
...
+def gaussexptail_func(x, mu, sigma, alpha, n):
+    t = (x - mu) / sigma * tf.sign(alpha)
+    abs_alpha = znp.abs(alpha)
+    cond = tf.less(t, -abs_alpha)
+    func = z.safe_where(
+        cond,
+        lambda t: znp.exp(-0.5 * znp.square(abs_alpha) + n * (abs_alpha + t)),
+        lambda t: znp.exp(-0.5 * znp.square(t)),
+        values=t,
+        value_safer=lambda t: znp.ones_like(t) * (abs_alpha - 2),
```
ikrommyd, line 563: @jonas-eschle what would be the proper `value_safer` here?

zfit/models/physics.py
jonas-eschle: ah yes, I misread! All fine

zfit/models/physics.py
```diff
@@ -548,3 +548,400 @@ class GeneralizedCBPDFRepr(BasePDFRepr):
...
+    alphar: Serializer.types.ParamTypeDiscriminated
+    nr: Serializer.types.ParamTypeDiscriminated
+
+
+GeneralizedGaussExpTail.register_analytic_integral(
+    func=generalized_gaussexptail_integral, limits=gaussexptail_integral_limits
+)
+
+
+class CBExGauss:
```
jonas-eschle, line 866: This requires some more thoughts I think. I would tendentially not follow the naming scheme and behavior used in the root implementation: "tailleft" is not very explanatory, the choices are highly arbitrary: why is the left tail GaussExp, and if it's a right one a CB?

We should probably start with: what does a user need? Maybe a one-tailed PDF with two sigmas, could be. In this case, it makes sense to be able to deactivate one tail, one _could_ know or read in the docs how to do it, i.e. set a high value, but if we do it, could also work.

If they want to mix both tales, that's another task (but is that needed?), and maybe a piecewise PDF could do the job, a more generalized one. Do you think the mixing of tales is needed?

For leaving away a tail, what about we allow the user to set parameters to `False`, i.e. either both left or right tail parameters in the generalized versions? This should be unambiguous. We could either set the limits high, but that's maybe a bit surprising for a user that the param is there, or simply not set them and adjust the (integral-)functions for the case that these parameters are not present in `params`.

Or what do you think, any specific use-cases in mind?

We could also just do nothing and have it in the docs?

jonas-eschle (commented):
Looks good overall, (still need a final go through once all is settled), there is surely some thoughts needed with the generalization, i.e. the CBEx factory.
Otherwise, we can also go without the factory first, both fine with me. Let's see if we find a good way

jonas-eschle:
ah, by chance, do you know other people who use this shapes and maybe have opinions on what is needed? Feel free to tag them and bring the along to the discussion, can only be helpful

zfit/models/physics.py
ikrommyd: Yes, actually let's not do the factory for now. Can I leave it commented out for future reference?

jonas-eschle: Sure, just leave a comment why it has been commented out, but that's fine

ikrommyd:
@jonas-eschle I've kept the `n` in the numpy functions in the tests to 1) check that `n` was removed correctly and 2) test that thet are equivalent for `n=alpha`. I would like to keep the tests like that.

tests/test_pdf_gaussexptail.py
```diff
@@ -0,0 +1,184 @@
...
+
+    rnd_limits = sorted(list(np.random.uniform(*bounds, 13)) + list(bounds))
+    integrals = [
+        gaussexptail.integrate((low, up), norm=False)
+        for low, up in zip(rnd_limits[:-1], rnd_limits[1:])
+    ]
+
+    integral = np.sum(integrals)
+    integral_full = gaussexptail.integrate(bounds, norm=False).numpy()
+    assert pytest.approx(integral_full.item()) == integral.item()
```
jonas-eschle, line 94: what is this `.item()` exactly and why is it needed? It should work without, right?

zfit/models/physics.py
```diff
@@ -548,3 +548,299 @@ class GeneralizedCBPDFRepr(BasePDFRepr):
...
+    abs_alpha = znp.abs(alpha)
+    cond = tf.less(t, -abs_alpha)
+    func = z.safe_where(
+        cond,
+        lambda t: znp.exp(0.5 * znp.square(abs_alpha) + abs_alpha * t),
+        lambda t: znp.exp(-0.5 * znp.square(t)),
+        values=t,
+        value_safer=lambda t: znp.ones_like(t) * abs_alpha,
+    )
+    return znp.maximum(func, znp.zeros_like(func))
```
jonas-eschle, line 565: two refactoring ideas:
- we can move the znp.exp outside, after the ` safe_where`?
- why do we need the znp.maximum? exp(something) is always positive?

zfit/models/physics.py
```diff
@@ -548,3 +548,299 @@ class GeneralizedCBPDFRepr(BasePDFRepr):
...
+            & \\mbox{for }0 \\leqslant \\frac{x - \\mu}{\\sigma_{R}} \\leqslant \\alpha_{R} \\newline
+            \\exp{\\left(\\frac{|\\alpha_{R}|^2}{2} - |\\alpha_{R}|  \\left(\\frac{x - \\mu}{\\sigma_{R}}\\right)\\right)},
+             & \\mbox{for }\\frac{x - \\mu}{\\sigma_{R}} > \\alpha_{R}
+            \\end{cases}
+
+        Args:
+            mu: The mean of the gaussian
+            sigmal: Standard deviation of the gaussian on the left side
+            alphal: parameter where to switch from a gaussian to the expontial tail on the left side
+            nl: Coefficient of the exponent in the exponential tail on the left side
```
jonas-eschle, line 782:
```suggestion
```

zfit/models/physics.py
```diff
@@ -548,3 +548,299 @@ class GeneralizedCBPDFRepr(BasePDFRepr):
...
+            sigmar: Standard deviation of the gaussian on the right side
+            alphar: parameter where to switch from a gaussian to the expontial tail on the right side
+            nr: Coefficient of the exponent in the exponential tail on the right side
```
jonas-eschle, line 785:
```suggestion
```

zfit/models/physics.py
```diff
@@ -548,3 +548,299 @@ class GeneralizedCBPDFRepr(BasePDFRepr):
...
+            obs: |@doc:pdf.init.obs| Observables of the
+               model. This will be used as the default space of the PDF and,
+               if not given explicitly, as the normalization range.
+
+               The default space is used for example in the sample method: if no
+               sampling limits are given, the default space is used.
+
+               The observables are not equal to the domain as it does not restrict or
+               truncate the model outside this range. |@docend:pdf.init.obs|
+            extended: |@doc:pdf.init.extended| The overall yield of the PDF.
```
jonas-eschle, line 795: just fyi, your'e aware how these work? It's a substitution, so all that is needed to type is `extended:|@doc:pdf.init.extended||@docend:pdf.init.extended|`
and it will automatically in the pre-commit fill the content from a file (in util/api)

jonas-eschle (changes requested):
Looks very good overall! Minor comments

zfit/models/physics.py
ikrommyd: Github copilot does that for me during editing haha. I didn't know.

tests/test_pdf_gaussexptail.py
ikrommyd: `.item()` in this case gets the element from an array with just 1 element. https://numpy.org/doc/stable/reference/generated/numpy.ndarray.item.html. I think without it, `pytest.approx` doesn't like the shapes.

zfit/models/physics.py
ikrommyd: I stole it from the crystal ball implementation. I can remove it

tests/test_pdf_gaussexptail.py
ikrommyd: Ok, it's fine in this case. I'm removing it. I've had cases when we're comparing a zfit result with a numba_stats result using `pytest.approx` where this is required.

jonas-eschle (approved):
LGTM!

PR #535, opened 2024-03-26
feat: add `QGauss` distribution

ikrommyd:
Adds `QGauss` distribution

## Proposed Changes

- Implement q-Gaussian for 1 < q < 3 by leveraging its relation to the Student's t distribution.
## Tests added

- Modify `tests/serialize/test_basics.py`  and add `tests/test_pdf_qgauss.py`

## Checklist

- [x] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated


zfit/models/dist_tfp.py
```diff
@@ -733,7 +733,62 @@ def __init__(
...
-        """Q-Gaussian distribution with parameter `q`."""
+        """Q-Gaussian distribution with parameter `q`.
+
+        The q-Gaussian is a probability distribution arising from the maximization of the Tsallis entropy under appropriate constraints.
+        It is defined for q < 3 and the Gaussian distribution is recovered as q -> 1.
+        For q < 1, is it the PDF of a bounded random variable.
+        We only support 1 < q < 3 in this implementation.
+        If you want to use exactly q = 1, use the `zfit.pdf.Gauss` class.
+        During fitting, if you want to start from a Gaussian shape, you can initialize the `q` parameter to be really close to 1.
+        It is related to the Student's t-distribution according to https://en.wikipedia.org/wiki/Q-Gaussian_distribution#Student's_t-distribution
```
jonas-eschle, line 744:
```suggestion
        It is related to the Student's t-distribution according to the `corresponding Wikipedia entry <https://en.wikipedia.org/wiki/Q-Gaussian_distribution#Student's_t-distribution>`_
```

tests/test_pdf_qgauss.py
```diff
@@ -0,0 +1,60 @@
...
+import pytest
+import zfit
+from zfit import z
+import zfit.z.numpy as znp
+from numba_stats import qgaussian as qgaussian_numba
+import tensorflow as tf
+
+mu_true = 0.5
+sigma_true = 1.1
+true_integral_dict = {1.00001: 0.9999782, 1.5: 0.9695, 2.0: 0.8063152, 2.5: 0.4805167, 2.9: 0.1078687, 2.99: 0.011044}
```
jonas-eschle, line 12: How was this obtained? Maybe a comment, in case things change? I assume numerically?

tests/test_pdf_qgauss.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+    np.testing.assert_allclose(
+        qgauss.pdf(tf.range(-5, 5, 10_000), norm=False).numpy(),
+        qgaussian_numba.pdf(tf.range(-5, 5, 10_000), q=q, mu=mu_true, sigma=sigma_true),
+        rtol=1e-5,
+    )
+    assert qgauss.pdf(tf.range(-5, 5, 10_000), norm=False) <= qgauss.pdf(0.5, norm=False)
+
+    sample = qgauss.sample(1000)
+    tf.debugging.assert_all_finite(sample.value(), "Some samples from the qgauss PDF are NaN or infinite")
```
jonas-eschle, line 35: Let's rather use `assert all(np.isfinite)`, I think that the tf doesn't really create an assert assertion but throws an error. Pytest testing should use an "assert" to best recognize tests and failures

zfit/models/dist_tfp.py
```diff
@@ -714,3 +717,115 @@ class StudentTPDFRepr(BasePDFRepr):
...
+
+        # https://en.wikipedia.org/wiki/Q-Gaussian_distribution
+        # relation to Student's t-distribution
+
+        # 1/(2 sigma^2) = 1 / (3 - q)
+        # 2 sigma^2 = 3 - q
+        # sigma = sqrt((3 - q)/2)
+
+        def dist_params():
+            df = (3 - q.value()) / (q.value() - 1)
```
jonas-eschle, line 809: We can add here a check if q is actually within the range using `tf.debugging.assert_*`, but only if `run.numeric_checks` is true.
you can import run locally
`from zfit import run`
and then use the tf check

zfit/models/dist_tfp.py
```diff
@@ -714,3 +717,115 @@ class StudentTPDFRepr(BasePDFRepr):
...
+        params = OrderedDict((("q", q), ("mu", mu), ("sigma", sigma)))
+
+        # https://en.wikipedia.org/wiki/Q-Gaussian_distribution
+        # relation to Student's t-distribution
+
+        # 1/(2 sigma^2) = 1 / (3 - q)
+        # 2 sigma^2 = 3 - q
+        # sigma = sqrt((3 - q)/2)
+
+        def dist_params():
```
jonas-eschle, line 808:
```suggestion
        def dist_params(q=q, mu=mu, sigma=sigma):
```
We can "capture" them: this way, the q is really q (by default, but we will never change it), i.e. by _reference_. The danger by not using this pattern is that `q` is only saved inside the function as a _lexical_, i.e. as a name.  For example, if later on in the init we would change `q` to something else, _that's_ what the function will use inside (it's gonna be a lexical lookup instead of by reference)

(yes, this isn't done with others (yet), but would be better if it were)

zfit/models/dist_tfp.py
```diff
@@ -714,3 +717,115 @@ class StudentTPDFRepr(BasePDFRepr):
...
+               Has no programmatical functional purpose as identification. |@docend:model.init.name|
+        """
+        q, mu, sigma = self._check_input_params(q, mu, sigma)
+        if q < 1 or q > 3:
+            msg = "q < 1 or q > 3 are not supported"
+            raise ValueError(msg)
+        if q == 1:
+            msg = "q = 1 is a Gaussian, use Gauss instead."
+            raise ValueError(msg)
+        params = OrderedDict((("q", q), ("mu", mu), ("sigma", sigma)))
```
jonas-eschle, line 799: you can use an ordinary dict (changed that for others in a PR, to be merged)

tests/test_pdf_qgauss.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.mark.parametrize("q", [1.00001, 1.5, 2.0, 2.5, 2.9, 2.99])
+def test_qgauss_pdf(q):
+    qgauss, _ = create_qgauss(q=q, mu=mu_true, sigma=sigma_true, limits=(-5, 5))
+    assert qgauss.pdf(0.5, norm=False).numpy().item() == pytest.approx(
+        qgaussian_numba.pdf(0.5, q=q, mu=mu_true, sigma=sigma_true), rel=1e-5
+    )
+    np.testing.assert_allclose(
+        qgauss.pdf(tf.range(-5, 5, 10_000), norm=False).numpy(),
```
jonas-eschle, line 28: maybe just factor out the `tf.range` calls as `testdata` or similar? just reduces the number of calls ( not a big game changer oc)

jonas-eschle (changes requested):
Good as usual, only minor comments again to polish it

zfit/models/dist_tfp.py
ikrommyd: changed it to
```python
        def dist_params(q=q, mu=mu, sigma=sigma):
            from zfit import run
            if run.numeric_checks:
                tf.debugging.assert_greater(q, z.numpy.asarray(1.0), message="q must be > 1")
                tf.debugging.assert_less(q, z.numpy.asarray(3.0), message="q must be < 3")
            df = (3 - q.value()) / (q.value() - 1)
            scale = sigma.value() / tf.sqrt(0.5 * (3 - q.value()))
            return {"df": df, "loc": mu.value(), "scale": scale}
```

zfit/models/dist_tfp.py
```diff
@@ -714,3 +718,120 @@ class StudentTPDFRepr(BasePDFRepr):
...
+               ``ext_*`` methods and the ``counts`` (for binned PDFs). |@docend:pdf.init.extended|
+            norm: |@doc:pdf.init.norm| Normalization of the PDF.
+               By default, this is the same as the default space of the PDF. |@docend:pdf.init.norm|
+            name: |@doc:model.init.name| Human-readable name
+               or label of
+               the PDF for better identification.
+               Has no programmatical functional purpose as identification. |@docend:model.init.name|
+        """
+        q, mu, sigma = self._check_input_params(q, mu, sigma)
+        if q < 1 or q > 3:
```
jonas-eschle, line 794: Sorry, somehow my comment previously here disappeared (maybe I forgot to send): we should also test with tf.debugging.assert, at least once it's not executing eagerly.
```
if run.executing_eagerly():
    if q < 1 or q > 3:...
    # the rest
elif run.numeric_checks:
    tf.debugging.assert...
    ```
The reason is that if this is created _inside_ a TF function, it will otherwise fail. I think python checks are in general better, the error is nicer etc, but if it happens inside a tf.function, the check will fail (i.e. it will error, because `q` is a symbolic tensor and doesn't have a truth value while Python runs)

jonas-eschle (approved):
Looks good, let's get this in

PR #538, opened 2024-03-28
feat: add `BifurGauss` PDF

ikrommyd:
Adds `BifurGauss` distribution
We choose this name to be consistent with ROOT and because it also fits the zfit naming standards.
Related to #512 

## Proposed Changes

- Implement  Bifurcated Gaussian (also called Split-Normal or Two-Piece Normal)

## Tests added

- Modify `tests/serialize/test_basics.py`  and add `tests/test_pdf_bifurgauss.py`

## Checklist

- [x] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated


jonas-eschle (approved):
Perfect, LGTM!

Issue #539, opened 2024-03-29
Voigt profile raises CannotConvertToNumpyError when using extended PDF


fabnap:
**Bugs** Please fill the below form _if possible_.

When trying to fit the Voigt profile (zfit.pdf.Voigt) using zfit.loss.ExtendedBinnedNLL, zfit raises CannotConvertToNumpyError

## Current Behaviour
simple example:
```
import zfit 
import numpy as np

normalnp = np.random.normal(loc=2,scale=3, size=10000)

obs = zfit.Space('x', limits=(-2, 5))
data_nobin = zfit.Data.from_numpy(obs=obs, array=normalnp)
binning = zfit.binned.RegularBinning(20, -2, 5, name="x")
energy_bin = zfit.Space("x", binning=binning)
data = data_nobin.to_binned(energy_bin)


mu = zfit.Parameter("mu", 1.2,-4,6)
sigma = zfit.Parameter("sigma", 1.3, 0.1, 10)
gamma = zfit.Parameter("gamma", 1.3, 0.1, 10)
voigt = zfit.pdf.Voigt(m=mu, sigma=sigma, gamma=gamma ,obs=obs)
voigt_yield = zfit.Parameter("voigt_yield", 1000, 0, 10000)
voigt_extended = zfit.pdf.BinnedFromUnbinnedPDF(voigt, space=energy_bin, extended=voigt_yield)
gauss = zfit.pdf.Gauss(mu=mu, sigma=sigma, obs=obs)
gauss_yield = zfit.Parameter("gauss_yield", 1000, 0, 10000)
gauss_extended = zfit.pdf.BinnedFromUnbinnedPDF(gauss, space=energy_bin, extended=gauss_yield)


nll = zfit.loss.ExtendedBinnedNLL(model=voigt_extended, data=data) # raises the CannotConvertToNumpyError
# nll = zfit.loss.ExtendedBinnedNLL(model=gauss_extended, data=data) # works fine
# nll = zfit.loss.ExtendedBinnedChi2(model=voigt_extended, data=data) # raises the CannotConvertToNumpyError
# nll = zfit.loss.ExtendedBinnedChi2(model=gauss_extended, data=data) # works fine

minimizer = zfit.minimize.Minuit()
# minimizer = zfit.minimize.ScipyLBFGSBV1()
# minimizer = zfit.minimize.NLoptLBFGSV1()

result = minimizer.minimize(nll)
print(result)

```

## Context (Environment)

[...]

PR #541, opened 2024-03-30
fix: remove bad analytic integral registration in Voigt

ikrommyd:
Fixes https://github.com/zfit/zfit/issues/539

### DIFF
--- a/zfit/models/basic.py
+++ b/zfit/models/basic.py
@@ -309,18 +309,6 @@ class Voigt(BasePDF, SerializableMixin):
         return znp.real(znp.faddeeva_humlicek(complex_z)) / (sigma * np.sqrt(2 * np.pi))
 
 
-def _voigt_integral_from_inf_to_inf(limits, params, model):
-    del model, limits  # unused, fixed limits
-    params["m"]
-    sigma = params["sigma"]
-    params["gamma"]
-    return sigma * np.sqrt(2 * np.pi)
-
-
-limits = Space(axes=0, limits=(-znp.inf, znp.inf))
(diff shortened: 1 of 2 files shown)


zfit/models/basic.py
```diff
@@ -309,18 +309,6 @@ def _unnormalized_pdf(self, x):
         return znp.real(znp.faddeeva_humlicek(complex_z)) / (sigma * np.sqrt(2 * np.pi))
```
jonas-eschle, line 311:
```suggestion
def _voigt_integral_from_inf_to_inf(limits, params, model):
    del model, limits  # unused, fixed limits
    params["m"]
    sigma = params["sigma"]
    params["gamma"]
    return sigma * np.sqrt(2 * np.pi)


limits = Space(axes=0, limits=(-znp.inf, znp.inf))
# todo: this only works if executing eagerly, which fails for at least the binned PDFs
# possible solution comes with the `space.static` concept (?) and a StaticLimitsNotAvailable error,
# falling back to whatever is available
#  Voigt.register_analytic_integral(func=_voigt_integral_from_inf_to_inf, limits=limits)
```

jonas-eschle (approved):
Seems reasonable, one suggestion to kep the code there but instead uncomment and leave the todo. Other than that, it's fine

zfit/models/basic.py
jonas-eschle: @ikrommyd just commited s.t. the ci runs, feel free to merge

jonas-eschle (approved):

jonas-eschle:
CI should be fixed with https://github.com/scikit-hep/mplhep/pull/488, merging as the changes are minimal and it worked before and runs locally

PR #548, opened 2024-04-05
feat: add `Gamma` distribution

ikrommyd:
Adds `Gamma` distribution
We choose the same parametrization [with ROOT](https://root.cern/doc/master/classRooGamma.html)
Related to #512 

## Proposed Changes

- Implement  Gamma distribution

## Tests added

- Modify `tests/serialize/test_basics.py`  and add `tests/test_pdf_pdf.py`

## Checklist

- [ ] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated


zfit/models/dist_tfp.py
```diff
@@ -933,3 +933,105 @@ class BifurGaussPDFRepr(BasePDFRepr):
...
+        obs: ztyping.ObsTypeInput,
+        *,
+        extended: ExtendedInputType = None,
+        norm: NormInputType = None,
+        name: str = "Gamma",
+    ):
+        """Gamma distribution.
+
+        The gamma shape is parametrized here with `gamma`, `beta` and `mu`.
+        We follow the same parametrization `as RooFit <https://root.cern.ch/doc/master/classRooGamma.html>`_.
```
jonas-eschle, line 955:
```suggestion
        The gamma shape is parametrized here with `gamma`, `beta` and `mu`, following
        the same parametrization `as RooFit <https://root.cern.ch/doc/master/classRooGamma.html>`_.
```

zfit/models/dist_tfp.py
```diff
@@ -933,3 +933,105 @@ class BifurGaussPDFRepr(BasePDFRepr):
...
+
+            f(x \\mid \\gamma, \\beta, \\mu) = (x - \\mu)^{\\gamma - 1} \\exp{\\left(-\\frac{x - \\mu}{\\beta}\\right)} / Z
+
+        with the normalization over [0, inf] of
+
+        .. math::
+
+            Z = \\Gamma(\\gamma) \\beta^{\\gamma}
+
+        The normalization changes for different normalization ranges
```
jonas-eschle, line 968:
```suggestion
        The normalization changes for different normalization ranges and `Z=1` for the unnormalized shape. 
```

zfit/models/dist_tfp.py
```diff
@@ -933,3 +933,105 @@ class BifurGaussPDFRepr(BasePDFRepr):
...
+            dist_params=dist_params,
+            obs=obs,
+            params=params,
+            name=name,
+            extended=extended,
+            norm=norm,
+        )
+
+    @property
+    def distribution(self):
```
jonas-eschle, line 1016: I would suggest to refactor this and remove the custom `distribution` property: above, `distribution` can be any callable that takes the parameters and returns a distribution, so we can just define this above in the init, approx
```
def distribution(mu, sigma,...):
    return TransformedDist(......)
```

(reduces duplicated code and streamlines the pdf)

jonas-eschle (changes requested):
Good overall, one refactoring and minor docs

zfit/models/dist_tfp.py
ikrommyd: Please take a look and let me know if this is what you were aiming for

jonas-eschle (approved):
Perfect, that's what I had in mind, looks good! LGTM

PR #551, opened 2024-04-10
use result to update params

jonas-eschle:
## Proposed Changes

- use a `FitResult` to update the parameters temporarily
- enable experimentally that the fit does _not_ update the default parameters

## Tests added

- to test the effects

## Checklist

- [x] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated

### DIFF
--- a/zfit/core/parameter.py
+++ b/zfit/core/parameter.py
@@ -1610,7 +1610,7 @@ def check_convert_param_values_assign(params, values, allow_partial=False):
     """
     if isinstance(params, ZfitResult) and values is None:
         params, values = None, params
-    else:
+    elif not isinstance(values, ZfitResult):
         params = convert_to_container(params)
         if params is None:
             msg = "No parameters given to set values to (values={values})."
--- a/zfit/core/parameter.py
+++ b/zfit/core/parameter.py
@@ -1615,8 +1615,9 @@ def check_convert_param_values_assign(params, values, allow_partial=False):
         if params is None:
             msg = "No parameters given to set values to (values={values})."
             raise ValueError(msg)
-        if len(params) == 0 or len(values) == 0:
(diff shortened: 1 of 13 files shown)


zfit/core/parameter.py
```diff
@@ -1576,7 +1578,14 @@ def set_values(
     """
     if allow_partial is None:
         allow_partial = False
-    params, values = check_convert_param_values_assign(params, values, allow_partial)
+    params, values, is_empty = check_convert_param_values_assign(params, values, allow_partial)
+    if is_empty:
+
+        @contextlib.contextmanager
+        def empty_context():
```
jonas-eschle, line 1585: We can't  do that, params also need to be reset

zfit/minimizers/baseminimizer.py
```diff
@@ -561,7 +561,16 @@ def _make_stateful(
         """
         state = {"loss": loss, "params": params, "init": init}
         self._state = state
+        from zfit import settings
+
+        if no_update := not settings.options.auto_update_params:
```
jonas-eschle, line 566: It's  true, right?

zfit/minimizers/fitresult.py
```diff
@@ -1526,6 +1486,30 @@ def _repr_pretty_(self, p, cycle):
             return
         p.text(self.__str__())
 
+    def __enter__(self):
+        self._tmp_old_param_values.append(znp.asarray(tuple(self.params.keys())))
+        self.update_params()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        old_vals = self._tmp_old_param_values.pop()
```
jonas-eschle, line 1495: Pop last one!?

PR #553, opened 2024-04-12
docs: add physics pdfs

jonas-eschle:
Signed-off-by: Jonas Eschle <mayou36@jonas.eschle.com>Fixes #

## Proposed Changes

- add the physics pdfs to docs, install zfit-physics to have access

## Checklist

- [x] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated


docs/user_api/zfit.pdf.rst
```diff
@@ -78,3 +78,15 @@ They offer more or less support and freedom.
     :maxdepth: 2
 
     pdf/custom_base
+
+Physics PDFs
```
jonas-eschle, line 82: @ikrommyd FYI, these are the parts of adding docs: I've added here a toctree that references a file above, the physics.rst situated in the folder "pdf", which is also created.

docs/user_api/pdf/physics.rst
```diff
@@ -0,0 +1,12 @@
+
+
+.. autosummary::
+    :toctree: _generated/physics
+
+    zfit_physics.pdf.Argus
+    zfit_physics.pdf.RelativisticBreitWigner
```
jonas-eschle, line 7: again @ikrommyd FYI: the autosummary directive creates a summary of objects that are listed, the toctree will be generated in `_generated`/physics -> `_generated` is just a folder, conventional, and physics the name of the file to be produced (which can be arbitrary, but we take the same as the rst file here).

PR #575, opened 2024-07-24
feat: Adding Levenberg Marquardt Optimizer

ConnorStoneAstro:
This PR is to add the LM optimizer to zfit. LM basically assumes local linearity by extracting the gradient (`g`) and Hessian (`H`), using a scalar damping parameter (`L`) it interpolates between gradient descent and Gauss-Newton optimization by solving an equation of the form `(H + LI)h = g` for a step (`h`) in parameter space which minimizes the loss function (`I` is the identity).

## Proposed Changes

- Adding the `zfit/minimizers/minimizer_lm.py` file which implements the LM algorithm

This PR is a work in progress, I believe it will work best as a collaborative effort. I have sketched out a single mode for the LM algorithm, though I am not familiar with the zfit system or TensorFlow. Thus I hope you can make changes to this PR to sync it up with the rest of the formalism. Then I'll implement the other modes following the given format. Then we can work on tests to ensure everything is working as expected. And maybe also eventually add a tutorial.

## Tests added

- TBD, this is a work in progress

## Checklist

- [x] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated


ikrommyd:
Haven't checked anything specific in the algorithm yet but one general comment I have is: Please use `tensorflow` and in particular `znp` like `import zfit.z.numpy as znp` which is almost an alias for the tensorflow numpy api [tf.experimental.numpy](https://www.tensorflow.org/api_docs/python/tf/experimental/numpy). We wouldn't want `torch` in there for instance as the backend is purely relying on tensorflow to do all the work. So replacing `nuppy` and `torch` functionality with `tensorflow` and preferably `znp` functionality is the first step.

ConnorStoneAstro:
Hi @ikrommyd , indeed this still needs a lot of work. I just put this together after meeting with @jonas-eschle at SciPy2024. I made this draft PR to get input right from the start, and also to ask for some help. I am not a `TensorFlow` or `znp` guru, so I was hoping to make this PR collaboratively. If the `znp` import behaves like `np` then I can make some more progress. But you'll see that I will still need help because I have almost certainly got the `zfit` parts wrong as well.

ikrommyd:
Yeah `znp` is exactly like the tensorflow numpy api. One good thing to start with as well is to write up a test that this minimizer just minimizes a scalar function as well.

jonas-eschle:
Great, thanks for making the PR! Yes indeed, we talked about this, and looks good as a first draft, has everything that is needed!

And indeed, as @ikrommyd also said, it's straightforward, `znp` should have the usual numpy namespace, otherwise `tf` has it, a five minute thing to replace (luckily, these days, they're very compatible, the frameworks!).

Yes on the collaborative effort, absolutely, if you want to make the changes and just try (there are unittest where the minimizers are also tested, you could just add it and propagate to try, or use any of the examples), please just go ahead, otherwise I can do it (should be able to do by the end of the week!) (or @ikrommyd in case you're interested to play around? for context, we met at SciPy and he is doing some astro fitting, and we figured that zfit minimizers can be used, but that also his LM implementation is pretty damn beefy :D)

jonas-eschle:
@ConnorStoneAstro thanks again for the PR! I've changed a few things, and now it's running, but the logic is probably flawed. I tried the `simple_fit.py` in the `examples` section.

If I didn't screw up the logic with shapes, it's going, I fixed a few things (minus sign for the gradient etc), but I've also added a few adhoc values.

It seems that L is currently always  a huge number, 1e+9, and doesn't change, although the loss actually seems to get minimized. I think the problem is rather in the logic of the minimizer now, maybe you know where things go wrong?

ConnorStoneAstro:
@jonas-eschle great! Thanks for getting things online. I'll get the logic fixed up when I get a chance. I have a few things that are very urgent, so it will be at least a week before I can get to this. But I'm very excited to see it all working!

ConnorStoneAstro:
@jonas-eschle I think I figured out one reason why the logic seems off for the LM optimizer. It seems the `evaluator` is using `maxiter` as it's limit before throwing a `zfit.util.exception.MaximumIterationReached`. Since a single iteration of LM may use many function calls, it helps to separate things into two `maxiter` and `maxfunevals` variables (scipy minimize also tracks these separately). I can bypass it for now, so I'll keep  working on the logic, but it's something you might want to consider in general for all the minimizers.

ConnorStoneAstro:
@jonas-eschle Ok I think I have the logic working. It seems to be minimizing the function in the `simple_fit.py` example you mentioned. I don't think the information is being passed to the `FitResult` object correctly though. It looks like the `params` is supposed to be an object, not just the tensor that I'm passing. It may be necessary to change how I track the `params` I tried using `assign_values(params, val)` but it looks like `params` is actually a list, so I'm not sure what the protocol is. If you could ensure the `FitResult` is being built correctly at the end then we are in good shape to start testing!

jonas-eschle:
I've updated the branch and made some change (pull before push), what I am running into now is the "OptimizeStop", which stops the iteration and therefore doesn't check if it actually converged.

I've also added a check with the convergenge criterion; this is there to make sure that minimizers stop at the same point.

(the CI fails for unrelated reasons, something with the instalation)

ConnorStoneAstro:
Hi @jonas-eschle it looks like the issue comes from the way parameters are being updated. I'm not really sure how the parameter object works. In the main loop I use `assign_values(params, params + step[0][:, 0])` and in the individual step I use `params1 = params + h[:, 0]`. In my mind these should produce the same thing, but they are giving different results even if `step[0]` and `h` have the same value. Could you Make sure the zfit parts are working properly? Alternatively, if you have a tutorial on how to use the parameter object, I would be able to fix it. I would like `param1` to be a temporary "proposal state" in the sub loop, but then in the main loop I would like to update `param` to the new value. I suspect things will work properly after that.

jonas-eschle:
Hi @ConnorStoneAstro I've pushed some changes. I am not sure exactly, what is happening: it seems to go great for the first step (you can use the `simple_fit` in the code examples section, and also change the verbosity down to say 7), but then goes to the wrong direction and gets stuck?

> I'm not really sure how the parameter object works. 

Basically, we never need to use the parameter objects, we can just use the values in an array form (for example to use the `value` or `gradient` function. It's just updated at the end and it should be working now (done a few hacks for trying around btw)

Do you see what is going wrong here?

jonas-eschle:
@ConnorStoneAstro good news! I've fixed it (typo...) and made a small comparison in the `simple_fit.py` (to be cleaned up again, just for hacking), and it seems to work well! Guess we finally start to have things together ;)

(I don't worry at all that for this example iminuit is "faster", I am sure this can greatly vary and I'd love to see more complex examples)

ConnorStoneAstro:
Awesome! Glad it's working 🙂 I can add the other optimization modes, and some unit tests next.

jonas-eschle:
@ConnorStoneAstro I've been doing quite some cleanup in zfit recently, I'll tackle this PR next and make it more mergeable, iterating here. My suggestion is to have it in and add other methods etc later, otherwise let me know!

ConnorStoneAstro:
@jonas-eschle sorry I've been so slow on this, great to hear about the updates! Let me know when you're done iterating on this PR

jonas-eschle:
@ConnorStoneAstro I've cleaned things up, added it to the tests, docs etc, I think this looks fine overall to me, what do you think?
The only thoughts are maybe on `rho_min` and `rho_max`, as we have also trust region optimizers, albeit not being the same, the names could maybe be optimized? Or even `rhomin` and `rhomax`? No strong opinion tough.

And agree on the number of function evaluations vs maxiter, this should be changed basically, but among all minimizers. So I propore to merge and then do it as a change everywhere

jonas-eschle (approved):
@ConnorStoneAstro  looks all good to me now, removed the draft status. If you're fine, I'll merge and get a new zfit release ou

ConnorStoneAstro:
Hi @jonas-eschle wow awesome! Yeah might as well merge and we can add more modes later. Looks like the unit tests aren't  passing, but I guess that's not related to this PR?

PR #599, opened 2024-11-08
feat: add `GeneralizedGauss` distribution

fchinu:
Adds `GeneralizedGauss` distribution

## Proposed Changes
- Implement `GeneralizedGauss` from [tensorflow probability](https://www.tensorflow.org/probability/api_docs/python/tfp/distributions/GeneralizedNormal)

## Tests added
- Modify `tests/serialize/test_basics.py` and add `tests/test_pdf_generalizedgauss.py`

## Checklist

- [x] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [x] CHANGELOG updated


jonas-eschle (approved):
I cannot find a single thing that I would change, @ikrommyd maybe you have something?

ikrommyd:
> I cannot find a single thing that I would change, @ikrommyd maybe you have something?
> 
> @fchinu feel free to change this from a draft to a real PR if you think it's ready

Yeah I looked at it briefly only from my phone yesterday. Looking at it better now, I don’t see anything I would change either. Looks good!

PR #660, opened 2025-08-05
docs: Correct Extended*NLL docstrings and argdocs.yaml

LarsHebenstiel:
## Proposed Changes

Update docstrings for Extended*NLL to the correct mathematics and to accurately reflect the calculation actually performed in code. In the case of `ExtendedBinnedNLL` previously documentation stated that it uses `\mathcal{poiss}(N_{modelbin_i}, N_{databin_i}) = N_{databin_i}^{N_{modelbin_i}} \frac{e^{- N_{databin_i}}}{N_{modelbin_i}!}` however in this docstring the `N_model` and `N_data` are backwards. The implementation of `_loss_func` for `ExtendedBinnedNLL` uses `poisson_loss_calc` defined [here](https://github.com/zfit/zfit/blob/2c9c6fd7f1f592baa85af0fcdd60887256974f7e/src/zfit/_loss/binnedloss.py#L51). That function calls to `tf.nn.log_poisson_loss` to evaluate actually `\mathcal{poiss}(N_{databin_i}, N_{modelbin_i}) = N_{modelbin_i}^{N_{databin_i}} \frac{e^{- N_{modelbin_i}}}{N_{databin_i}!}`. This confusion might be related to the `TODO` comment on [this line](https://github.com/zfit/zfit/blob/2c9c6fd7f1f592baa85af0fcdd60887256974f7e/src/zfit/_loss/binnedloss.py#L314) of the `ExtendedBinnedNLL`. If PR approved maybe we can also delete this comment. In `ExtendedUnbinnedNLL` the same calculation is done [here](https://github.com/zfit/zfit/blob/2c9c6fd7f1f592baa85af0fcdd60887256974f7e/src/zfit/core/loss.py#L1259) without any comments asking to check the order of model and data.

Likelihood is given `L(model | data) = p(data | model)`, so the order of `N_data` and `N_model` currently in the code is correct and I have corrected it in the docs. The new formula matches my expectation that `N_data` is strictly an integer and `N_model` may be any positive real number in the case of simple Poisson statistics. Of course in the scaled poisson distribution (SPD) for weighted counts `N_data` may also be a positive real. I see some comments in the code about SPD [here](https://github.com/zfit/zfit/blob/2c9c6fd7f1f592baa85af0fcdd60887256974f7e/src/zfit/_loss/binnedloss.py#L68-L69). Recently I wrote some toy examples working with weighted data and I also noticed that `zfit` is currently ignoring the Bohm-Zech SPD error correction term. Maybe I can look more into the currently disabled implementation in zfit?

## Checklist

- [] change approved
- [x] implementation finished
- [x]  docs updated
- [] correct namespace imported
- [] tests added
- [] CHANGELOG updated

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,32 @@ Changelog
 
 .. _newest-changelog:
 
+Develop
+======================
+
+Major Features and Improvements
+-------------------------------
+
+Breaking changes
+------------------
+
+Deprecations
+-------------
+
+Bug fixes and small changes
+---------------------------
+- ComposedParameter now ignores keyword-only arguments when deciding whether to unpack_params
+
+Experimental
+------------
+
+Requirement changes
+-------------------
+
+Thanks
+------
+
+
 0.27.0 (14 Jul 2025)
 ======================
 
--- a/src/zfit/core/parameter.py
+++ b/src/zfit/core/parameter.py
@@ -1086,7 +1086,12 @@ class ComposedParameter(SerializableMixin, BaseComposedParameter):
             raise BreakingAPIChangeError(msg)
         takes_no_params = False
         if unpack_params is None:
-            parameters = signature(func).parameters
+            parameters_with_keyword_only = signature(func).parameters.values()
+            parameters = {
+                parameter.name: parameter
+                for parameter in parameters_with_keyword_only
+                if parameter.kind != parameter.KEYWORD_ONLY
+            }
             if isinstance(params, ZfitParameter):
                 params = [params]
                 unpack_params = True
--- a/tests/test_parameter.py
+++ b/tests/test_parameter.py
@@ -1,6 +1,8 @@
 #  Copyright (c) 2025 zfit
 from __future__ import annotations
 
+import math
+
 import numpy as np
 import pytest
 import tensorflow as tf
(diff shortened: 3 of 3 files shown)


jonas-eschle (approved):
Thanks for this fix! And yes, we could think about what are better loss functions, i.e. should we use the weighted correction by Bohm?

LarsHebenstiel:
> should we use the weighted correction by Bohm?

[Here's](https://scikit-hep.org/iminuit/notebooks/weighted_histograms.html) some toy studies on the weighted correction in the iminuit docs. And [here](https://scikit-hep.org/iminuit/reference.html#iminuit.cost.BinnedNLL) their docs state they do use the Bohm-Zech transform for weighted data inputted to `BinnedNLL`. Their docs are not clear whether they use it for `ExtendedBinnedNLL` also, but I poked around in the source code and from what I can tell the transform is used for all binned data.

The improvement is that the Bohm-Zech transform matches also the second moment (variance) of the weighted counts rather than just the first moment (mean). This improvement can drastically impact the error estimation on model parameters. In cases for example where the weights for the data are all small (< 1) the Weight histogram will have variance < counts. Using a poisson model for the bins will overestimate the variance of the data and you end up with correspondingly too large error bars. On the other hand for cases where the weights are all large (> 1) the opposite problem happens and the poisson model underestimates variance in each bin and the parameter estimates end up having too small error bars.

I don't think I know about the transform to understand what the other side effects of using the weighted correction are. I would need to read/think about it in depth first. Generally it seems like a good transform to use and there might be some more information in the iminuit issues/PR history why they chose to use it.

PR #661, opened 2025-08-06
Change ComposedParameter behavior

### LINKED ISSUE #659: Unexpected behavior of math.prod in ComposedParameter
## Current Behaviour

I am trying to make a `ComposedParameter` as a product of an arbitrary number of parameters.
From the documentation I am trying `math.prod`, but when I run

```python
import math
import zfit

# Define two base parameters
a = zfit.Parameter("a", 2.0, 0.0, 5.0)
b = zfit.Parameter("b", 3.0, 0.0, 5.0)

# Define a CompositeParameter as the product of a and b
prod_1 = zfit.ComposedParameter("prod_1", lambda x: math.prod(x), params=[a, b])
prod_2 = zfit.ComposedParameter("prod_2",              math.prod, params=[a, b])

print(prod_1)
print(prod_2)
```

the first product prints as expected, while the second shows me some ERROR as in:

```
<zfit.ComposedParameter 'prod_1' params=[('param_0', 'a'), ('param_1', 'b')] value=6>
<zfit.ComposedParameter 'prod_2' params=[('param_0', 'a'), ('param_1', 'b')] value=ERROR OCCURRED>
```

## Expected Behaviour

I would expect both to be valid and I would rather use the second approach with only `math.prod`, given that it's shorter.

## Context (Environment)

- zfit version:  0.27.1
- Python version: 3.12.11
- Are you using conda, pipenv, etc? : micromamba
- Operating System: Almalinux9
- Tensorflow version: 2.19.0

LarsHebenstiel:
Fixes #659

## Proposed Changes

- ComposedParameter now ignores keyword-only arguments when automatically deciding whether to unpack internal parameter list to pass to function. This should not change behavior for already working use cases.

## Tests added

- None

## Checklist

- [ ] change approved
- [x] implementation finished
- [ ]  docs updated
- [ ] tests added
- [x] CHANGELOG updated


jonas-eschle (approved):

tests/test_parameter.py
```diff
@@ -595,3 +597,78 @@ def test_invalid_parameter_names():
...
+    prod_10 = zfit.ComposedParameter("prod_10", math.prod, params=[x, y, z]) # Should not unpack
+    with pytest.raises(ValueError, match="Cannot determine if parameter should be unpacked or not. Please specify explicitly `unpack_params`"):
+        prod_11 = zfit.ComposedParameter("prod_11", my_ambiguous_function, params=[x, y, z])
+    prod_12 = zfit.ComposedParameter("prod_12", my_ambiguous_function, params=[x, y, z], unpack_params=False)
+
+
+    assert prod_10.value() == 24
+    assert prod_12.value() == 24
+
+    pass
```
jonas-eschle, line 674:
```suggestion
```

jonas-eschle: Looks good to me, thanks!

jonas-eschle (approved):

jonas-eschle (approved):

PR #669, opened 2025-09-04
chore: update Python version to 3.13 in CI and configuration files

jonas-eschle:
Fixes #

## Proposed Changes

-

## Tests added

-

## Checklist

- [ ] change approved
- [ ] implementation finished
- [ ]  docs updated
- [ ] tests added
- [ ] CHANGELOG updated


jonas-eschle:
Fails due to some dependency issue, where it install llvmlite. Fixing numpy to < 2.3 works, but let's way for fix

PR #726, opened 2026-03-04
Fix variance semantics (σ²) in binned chi2 and Poisson auto-variances

### LINKED ISSUE #725: Inconsistent variances semantics (σ vs σ²) in binned Chi2 and BinnedData.from_tensor(variances=True) leads to wrong χ² weighting
**Bugs** 

Binned χ² calculations in zfit appear to mix up **standard deviation (σ)** and **variance (σ²)** in multiple places. In particular, `BinnedChi2`/`ExtendedBinnedChi2` with `errors="expected"` uses `sqrt(expected)` as the “variances” term while the χ² implementation divides by `variances` (treating it as σ²), leading to silently incorrect bin weighting. Additionally, `BinnedData.from_tensor(..., variances=True)` generates Poisson uncertainties consistent with σ (sqrt(values)) rather than Poisson variances (Var=N), which can also mis-weight χ² when `errors="data"` uses `data.variances()` in the denominator. The expected behavior is that `variances` consistently denotes **variance (σ²)** (UHI-compatible), with Poisson defaults `Var[N]=N` and `errors="expected"` using `Var=μ` in the denominator. Regression tests and a PR are available to reproduce and fix the issue.

Note: UHI defines `variances()` as variances of the bin contents (σ²), so aligning zfit `variances` to σ² is consistent with UHI.

## Current Behaviour

`variances` appears to be used inconsistently as **standard deviation (σ)** in some binned paths and as **variance (σ²)** in others.

1) `BinnedChi2` / `ExtendedBinnedChi2` with `options={"errors": "expected"}` constructs “variances” from the model expectation using `sqrt(expected)` (σ), but the χ² implementation divides by `variances` (treating it as σ²). This silently mis-weights χ² (i.e. uses Δ²/σ instead of Δ²/σ²).

2) `BinnedData.from_tensor(..., variances=True)` generates Poisson “uncertainties” consistent with σ (sqrt(values)) instead of Poisson **variance** (values). When used with `BinnedChi2(errors="data")` (which uses `data.variances()` as denominator), χ² values are mis-weighted.

```python
import zfit
import zfit.z.numpy as znp
from zfit._data.binneddatav1 import BinnedData
from zfit.models.template import BinnedTemplatePDFV1

obs = zfit.Space("x", limits=(0.0, 2.0), binning=2)
expected = znp.asarray([50.0, 80.0], dtype=znp.float64)
observed = znp.asarray([40.0, 90.0], dtype=znp.float64)

model_data = BinnedData.from_tensor(space=obs, values=expected)
model = BinnedTemplatePDFV1(data=model_data)

# Chi2 currently requires data variances to be present even for errors="expected"
data = BinnedData.from_tensor(space=obs, values=observed, variances=observed)

loss = zfit.loss.BinnedChi2(
    model=model,
    data=data,
    options={"errors": "expected", "empty": "ignore"},
)

# Pearson chi2 with Poisson variance: Var = N_exp
chi2_manual = float(znp.sum((expected -observed) ** 2 / expected))

# zfit returns a tensor-like; cast to float
chi2_val = float(loss.value())

print(abs(chi2_val - chi2_manual)/chi2_val)

## output : 0.8716555180966132. should be ~0.

```

## Expected Behaviour

Expected Pearson χ² for errors="expected" (Poisson Var=μ),

$\chi^2 = \sum\frac{(N-\mu)^2}{\mu}$ 

but current behavior corresponds to dividing by sqrt(μ) in the expected branch, giving a different value. 
The chi2_val is different from chi2_manual.

## Context (Environment)

This affects binned χ² fits because it silently changes bin weights and therefore can change fit results without raising an error (σ vs σ² mix-up), especially noticeable in low-statistics bins.

- zfit version: 0.28.1.dev79+g224d1a03f
- Python version: 3.11.11
- Are you using conda, pipenv, etc? : venv
- Operating System: ubuntu 24.04
- Tensorflow version: 2.20.0

## Possible Solution/Implementation
 
- Keep χ² denominator semantics as 1/variances (interpreting variances as σ²).

- Fix errors="expected" in BinnedChi2 and ExtendedBinnedChi2 to use expected variance: variances = expected + tiny instead of sqrt(expected + tiny).

- Align BinnedData.from_tensor(..., variances=True) with Poisson variance semantics: variances = values (for unweighted counts).

- Update docs wording (“uncertainties”) to explicitly state variances are σ² and Poisson default is Var[N]=N.

I have a PR with regression tests covering both chi2 classes and both error modes; happy to link it here once the issue is created.

YANLJX17:
Fixes #725

### Why
Current behavior can silently mis-weight χ² by mixing σ and σ², affecting fit results without raising errors.

## Proposed Changes

 - Align binned variances semantics to consistently represent variance (σ²) (UHI-compatible) across binned data construction and chi² losses.

 - Fix BinnedChi2 and ExtendedBinnedChi2 with options={"errors": "expected"} to use expected Poisson variance Var = μ as denominator (instead of sqrt(μ)).

 - Make BinnedData.from_tensor(..., variances=True) generate Poisson variances (for unweighted counts: Var[N] = N) instead of standard deviations.

 - (Docs) Clarify that variances denotes σ² and that Poisson default is Var[N]=N.

## Tests added

- Regression tests covering both BinnedChi2 and ExtendedBinnedChi2 for:

  - errors="expected": Pearson χ² uses expected variance (σ²=μ)
  - errors="data": χ² uses data.variances() as σ². 
(see tests/test_binnedloss.py)

- Regression tests covering BinnedData for:
  -  BinnedData.from_tensor, variances=True: treat variances=σ².

## Behavior change: 
- BinnedData.from_tensor(..., variances=True) now returns Poisson variance (N) rather than standard deviation (sqrt(N)), aligning with the variances naming/UHI semantics and χ² denominator usage.

Happy to add a CHANGELOG entry if you’d like this documented as a behavior change.

## Checklist

- [ ] change approved
- [x] implementation finished
- [x]  docs updated
- [x] tests added
- [ ] CHANGELOG updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ tmp/*
 /CLAUDE.md
 /.claude/
 /tmp/
+.venv/
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -20,6 +20,8 @@ Deprecations
 Bug fixes and small changes
 ---------------------------
 - Fix ``zfit.run.set_cpus_explicit()`` which was failing after ``import zfit`` due to premature TensorFlow initialization.
+- Fix ``BinnedChi2`` and ``ExtendedBinnedChi2`` to use expected Poisson variance in the ``errors="expected"`` path.
+- Make ``BinnedData.from_tensor(..., variances=True)`` return Poisson variances (``Var[N] = N``) for unweighted counts, aligning binned ``variances`` semantics with σ².
 
 Experimental
 ------------
--- a/src/zfit/_data/binneddatav1.py
+++ b/src/zfit/_data/binneddatav1.py
@@ -229,14 +229,14 @@ class BinnedData(
             values: |@doc:binneddata.param.values| Corresponds to the counts of the histogram.
                Follows the definition of the
                `Unified Histogram Interface (UHI) <https://uhi.readthedocs.io/en/latest/plotting.html#plotting>`_. |@docend:binneddata.param.values|
-            variances: |@doc:binneddata.param.variances| Corresponds to the uncertainties of the histogram.
+            variances: |@doc:binneddata.param.variances| Corresponds to the variances (sigma**2) of the histogram.
                If ``True``, the uncertainties are created assuming that ``values``
-               have been drawn from a Poisson distribution. Follows the definition of the
+               have been drawn from a Poisson distribution (i.e. ``Var[N] = N`` for unweighted counts). Follows the definition of the
                `Unified Histogram Interface (UHI) <https://uhi.readthedocs.io/en/latest/plotting.html#plotting>`_. |@docend:binneddata.param.variances|
         """
         values = znp.asarray(values, znp.float64)
         if variances is True:
-            variances = znp.sqrt(values)
+            variances = values 
         elif variances is not None:
             variances = znp.asarray(variances)
         return cls(
--- a/src/zfit/_data/binneddatav1.py
+++ b/src/zfit/_data/binneddatav1.py
@@ -229,9 +229,9 @@ class BinnedData(
             values: |@doc:binneddata.param.values| Corresponds to the counts of the histogram.
                Follows the definition of the
                `Unified Histogram Interface (UHI) <https://uhi.readthedocs.io/en/latest/plotting.html#plotting>`_. |@docend:binneddata.param.values|
-            variances: |@doc:binneddata.param.variances| Corresponds to the uncertainties of the histogram.
+            variances: |@doc:binneddata.param.variances| Corresponds to the variances (sigma**2) of the histogram.
                If ``True``, the uncertainties are created assuming that ``values``
-               have been drawn from a Poisson distribution. Follows the definition of the
+               have been drawn from a Poisson distribution (i.e. ``Var[N] = N`` for unweighted counts). Follows the definition of the
                `Unified Histogram Interface (UHI) <https://uhi.readthedocs.io/en/latest/plotting.html#plotting>`_. |@docend:binneddata.param.variances|
         """
         values = znp.asarray(values, znp.float64)
--- a/src/zfit/_data/binneddatav1.py
+++ b/src/zfit/_data/binneddatav1.py
@@ -229,14 +229,14 @@ class BinnedData(
             values: |@doc:binneddata.param.values| Corresponds to the counts of the histogram.
                Follows the definition of the
                `Unified Histogram Interface (UHI) <https://uhi.readthedocs.io/en/latest/plotting.html#plotting>`_. |@docend:binneddata.param.values|
-            variances: |@doc:binneddata.param.variances| Corresponds to the variances (sigma**2) of the histogram.
+            variances: |@doc:binneddata.param.variances| Corresponds to the uncertainties of the histogram.
                If ``True``, the uncertainties are created assuming that ``values``
-               have been drawn from a Poisson distribution (i.e. ``Var[N] = N`` for unweighted counts). Follows the definition of the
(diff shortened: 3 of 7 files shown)


jonas-eschle:
Fair enough, that's a good catch, thanks a lot! Can you add a corresponding CHANGELOG entry?

YANLJX17:
> Fair enough, that's a good catch, thanks a lot! Can you add a corresponding CHANGELOG entry?

Thanks! I’ve added the CHANGELOG entry and pushed the update. Marking this PR ready for review now.

jonas-eschle (approved):

jonas-eschle:
Thanks a lot for, I'll merge this later.

Out of curiosity, we're surveying a bit, how much of this did you write yourself and how much is AI generated? And how did you find it, yourself or AI?

Don't worry, even if everything is AI generated and found by AI, I'll merge it and it's still attributed to you: the PR looks good, plain and correct, that's all that matters.

YANLJX17:
> Thanks a lot for, I'll merge this later.
> 
> Out of curiosity, we're surveying a bit, how much of this did you write yourself and how much is AI generated? And how did you find it, yourself or AI?
> 
> Don't worry, even if everything is AI generated and found by AI, I'll merge it and it's still attributed to you: the PR looks good, plain and correct, that's all that matters.

  Hi, happy to talk about that. I am a PhD student from the CDEX group at Tsinghua University. My email address is liujiaxu21@mails.tsinghua.edu.cn. This is my first PR on GitHub.
  I did use AI (ChatGPT), but only for help with phrasing the problem description and for advice on the GitHub issue and PR submission process. I had actually discovered this issue myself a long time ago (around 2024), and it is quite an old issue. However, since I was not very familiar with the standard PR workflow, I put it aside. Only recently, when I started working on my project again, did I remember this issue. I then asked AI how to contribute a PR in a more professional way, and the result seems to have been quite good.

jonas-eschle:
Thanks, that makes sense.
In general, you could have also just opened an issue to report the problem instead of directly creating a PR (that's the default workflow, but this also works of course!).

The AI is perfectly fine. I use it on a daily basis too. The important thing is that you can sign of on every single letter and that it doesn't _look_ AI generated. That everything makes sense. Good use of AI!

jonas-eschle:
I see a unittest failing, can you run it? tests/data/test_binneddata.py::test_variance

Maybe the test is actually flawed, but it makes sense to fail, it probably has to do with the correct variance now

YANLJX17:
> I see a unittest failing, can you run it? tests/data/test_binneddata.py::test_variance
> 
> Maybe the test is actually flawed, but it makes sense to fail, it probably has to do with the correct variance now

Thanks, good catch. I ran `tests/data/test_binneddata.py::test_variance` locally — it was still asserting the previous `sqrt(values)` behavior for `variances=True`. Since `variances` now consistently denotes variance (σ²), I updated the test to check Poisson variance (`Var[N] = N`) instead, and pushed the fix.

jonas-eschle:
Perfect, makes sense!

YANLJX17:
> Thanks, that makes sense. In general, you could have also just opened an issue to report the problem instead of directly creating a PR (that's the default workflow, but this also works of course!).
> 
> The AI is perfectly fine. I use it on a daily basis too. The important thing is that you can sign of on every single letter and that it doesn't _look_ AI generated. That everything makes sense. Good use of AI!

Thanks! Just to clarify, I actually opened the issue first, then created the PR, and linked them afterwards so everything was cross-referenced properly. I’m not sure whether this matches your default workflow.

And thanks — I used AI mostly as a drafting aid, but I made sure to verify what the AI wrote locally.

PR #735, opened 2026-07-10
fix: compatibility with mplhep 1.3+ and ArviZ 1.0+, drop Python <3.12 support

raulgquartieri:
## Proposed Changes
- **Drop Python 3.10/3.11 support**: Bumped `requires-python` to `>=3.12`. This lets us drop the `typing_extensions.Self` compatibility shim (`sys.version_info >= (3, 11)` branches) in `core/basepdf.py`, `models/conditional.py`, and `models/special.py` in favor of `typing.Self` directly, and drop the now-unneeded `typing_extensions` dependency. CI workflow matrix updated from Python 3.10 to 3.12.

- **Dependency bumps**: `numpy>=1.16` → `>=2.1`, `arviz>=0.14.0` → `>=1.0.0`, `pytest>=3.4.2` → `>=8`. Dropped the `pytest-cases` dependency along with its last remaining usage in `tests/model/test_pdf_params_args.py`.

- **ArviZ 1.0 Compatibility**:
  - Updated the `to_arviz` conversion logic in `src/zfit/_bayesian/posterior.py` to wrap the parameter dict under a `"posterior"` key, matching the `az.from_dict` structure ArviZ 1.0+ expects.
  - Migrated `examples/bayesian_inference.py` from `arviz`'s `plot_trace`/`plot_posterior`/`plot_pair`/`plot_autocorr` to the new `arviz_plots` package equivalents, and updated `idata.groups()` → `idata.groups` (now a property, not a method, under ArviZ 1.0's DataTree-based `InferenceData`).

- **mplhep Compatibility**: Updated `zfit.data.BinnedData` (and `BinnedSamplerData`) to be UHI-compliant by adding a `flow` keyword argument to `values()` and `variances()`. This resolves the `TypeError` encountered when passing binned data to `mplhep` 1.3+.

- **Dynamic Padding**: Implemented dynamic padding using `tf.pad` for `flow=True` requests. This ensures that the returned bin counts correctly match UHI expectations (padding the bins to include underflow/overflow) without modifying the underlying tensor when `flow=False`.

## Tests added
None

## Checklist
- [ ] change approved
- [x] implementation finished
- [ ] docs updated
- [ ] tests added
- [x] CHANGELOG updated 

**Note on testing:**
I have verified the mplhep/ArviZ fix by running the full documentation build. This process previously failed due to compatibility issues with both ArviZ 1.0 and `mplhep` 1.3. With these changes, the documentation build now completes successfully.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -19,6 +19,8 @@ Deprecations
 
 Bug fixes and small changes
 ---------------------------
+- Fix ``mplhep`` compatibility: update ``BinnedData.values()`` and ``BinnedData.variances()`` to support the UHI ``flow`` keyword argument.
+- Fix ``ArviZ`` compatibility: update ``to_arviz`` conversion to align with 1.0+ API changes.
 - Fix ``zfit.run.set_cpus_explicit()`` which was failing after ``import zfit`` due to premature TensorFlow initialization.
 - Fix ``BinnedChi2`` and ``ExtendedBinnedChi2`` to use expected Poisson variance in the ``errors="expected"`` path.
 - Make ``BinnedData.from_tensor(..., variances=True)`` return Poisson variances (``Var[N] = N``) for unweighted counts, aligning binned ``variances`` semantics with σ².
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
       fail-fast: false
       matrix:
         use-graph: [ 1 ]
-        python-version: [ "3.10" ]  # omit intermediate versions to save compute time
+        python-version: [ "3.12" ]  # omit intermediate versions to save compute time
         os: [ ubuntu-latest ]
         extras: [ test ]
         run: [ true ]
@@ -205,7 +205,7 @@ jobs:
       fail-fast: False
       matrix:
         use-graph: [ 1 ]
-        python-version: [ "3.10" ]
+        python-version: [ "3.12" ]
         run: [ true ]
         extras: [ all ]
         include:
@@ -260,7 +260,7 @@ jobs:
       fail-fast: False
       matrix:
         use-graph: [ 1 ]
-        python-version: [ "3.10" ]
+        python-version: [ "3.12" ]
         run: [ true ]
         include:
           - python-version: "3.13"
--- a/examples/bayesian_inference.py
+++ b/examples/bayesian_inference.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 from pathlib import Path
 
 import arviz as az
+import arviz_plots
 
 # os.environ["CUDA_VISIBLE_DEVICES"] = "-1"  # disable GPU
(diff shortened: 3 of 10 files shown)


raulgquartieri:
@jonas-eschle - Apparently, using ArviZ > 1.0, as we discussed, breaks compatibility with Python 3.10 in the CI. How  we should proceed about this?

jonas-eschle:
Good point! Let's drop Python 3.10, it's not supported anymore by many libraries. You can just make this in the same PR, generally upgrading dependencies

raulgquartieri:
@jonas-eschle

src/zfit/_variables/axis.py
```diff
@@ -66,44 +68,139 @@ def __init__(self, name, lower=None, upper=None):
...
 
 
-# TODO: fill out below and don't just use the hist objects
+def _to_pyfloat(value) -> float:
+    """Convert a python number, numpy array/scalar or (eager) tensor with a single element to a python float."""
+    return float(np.asarray(value).reshape(-1)[0])
+
+
 class HashableAxisMixin:
-    def __init__(self, *args, **kwargs) -> None:
```
jonas-eschle, line 71: I am just a bit confused about these changes? Did you do them or are they AI gen? Perfectly fine as well, just a matter of how to review. What was the motivation for these changes?

CHANGELOG.rst
```diff
@@ -13,12 +13,16 @@ Major Features and Improvements
 
 Breaking changes
 ------------------
+- Drop support for Python 3.10 and 3.11; the minimum supported Python version is now 3.12.
+- ``zfit.binned.RegularBinning``/``VariableBinning`` no longer subclass ``hist.axis.Regular``/``Variable``, and ``Binnings`` no longer subclasses ``hist.axestuple.NamedAxesTuple``. They remain fully interoperable with ``hist``, ``boost_histogram``, and ``mplhep`` via duck-typing, but ``isinstance(x, hist.axis.Regular)`` (or ``hist.axestuple.NamedAxesTuple``) checks against zfit's binning objects no longer return ``True``.
```
jonas-eschle: Hm, so why exactly?

jonas-eschle (changes requested):

src/zfit/_variables/axis.py
raulgquartieri: This was AI gen. I was trying to resolve the "# TODO: fill out below and don't just use the hist objects" comment in the file. I've already reverted that change, so this PR is focused only on compatibility with mplhep 1.3+ and ArviZ 1.0+, and dropping Python <3.12 support, as was the original intent.

CHANGELOG.rst
raulgquartieri: I've already reversed this change.

raulgquartieri:
@jonas-eschle

raulgquartieri:
ping  @jonas-eschle

